PR #23, opened 2021-08-18
Feature auth providers buttons

romanetar:
New method formatAuthProviderButtons extracted from registration and registration-lite


smarcet:
please branch for master
regards

PR #24, opened 2021-08-18
Feature auth providers buttons format

romanetar:
New method formatAuthProviderButtons extracted from registration and registration-lite


smarcet (approved):
looks good to me

PR #34, opened 2022-02-08
Adding aria-labelledby prop to dropdown component

tomrndom:
* Adds required property to improve accesibility 

ref: https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=2791813&fileview=grid

### DIFF
--- a/src/components/inputs/dropdown.js
+++ b/src/components/inputs/dropdown.js
@@ -42,7 +42,7 @@ export default class Dropdown extends React.Component {
 
     render() {
 
-        let {onChange, value, className, error, clearable, disabled, overrideCSS, ...rest} = this.props;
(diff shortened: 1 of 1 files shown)


src/components/inputs/dropdown.js
```diff
@@ -42,7 +42,7 @@ export default class Dropdown extends React.Component {
 
     render() {
 
-        let {onChange, value, className, error, clearable, disabled, overrideCSS, ...rest} = this.props;
+        let {onChange, value, className, error, clearable, disabled, overrideCSS, ariaLabelledby, ...rest} = this.props;
```
smarcet, line 45: what if the user does not provides the attr ? i do think we should think about that case

smarcet (changes requested):
please review comments

PR #49, opened 2022-04-12
fix event types call query structure OR -> AND

romanetar:
ref https://tipit.avaza.com/task#!sortby=DateUpdatedDesc&task=2892481

Signed-off-by: romanetar <roman_ag@hotmail.com>


smarcet (approved):
LGTM

PR #78, opened 2022-08-30
Setting initial value from props on operator input

tomrndom:
ref: https://tipit.avaza.com/task#!sortby=DateUpdatedDesc&task=3008861


smarcet (approved):
LGTM

PR #81, opened 2022-09-05
Updates state when data from props is updated

tomrndom:
* Add missing key prop
* Updates states when data from props changes

ref: https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskProject&view=vertical&task=3033893&fileview=grid


smarcet (approved):
LGTM

PR #82, opened 2022-09-06
Moving dropItem function from drop to end drag

tomrndom:
* Fix the bug when the item is dropped outside the container but the order is changed

ref: https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskProject&view=vertical&task=3033893&fileview=grid


smarcet (approved):
LGTM

PR #91, opened 2022-10-18
Add new prop to trigger option reload of react-select component

tomrndom:
* Fix the bug that the dropdown shows options that are already on the table. The components re-renders but the dropdown isn't reloading their options. 
* Related to https://github.com/JedWatson/react-select/issues/1581#issuecomment-408625770

ref: https://tipit.avaza.com/task#!sortby=DateUpdatedDesc&task=3072749


smarcet (approved):
LGTM

PR #113, opened 2023-03-23
Add ...rest props to venue select component

tomrndom:
ref: https://tipit.avaza.com/task#!sortby=DateUpdatedDesc&task=3214506

### DIFF
--- a/src/components/schedule-builder-view/schedule-builder-view.less
+++ b/src/components/schedule-builder-view/schedule-builder-view.less
@@ -188,7 +188,7 @@
   z-index: @event_z_index + 2;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/schedule-builder-view/schedule-builder-view.less
```diff
@@ -188,7 +188,7 @@
   z-index: @event_z_index + 2;
 }
 
-.venues-selector, .day-selector, .event-type-selector {
+.day-selector, .event-type-selector {
```
smarcet, line 191: are we sure this does not break the schedule builder ?

smarcet (approved):
LGTM

PR #146, opened 2024-06-11
Set default value to avoid crash on char count initialization

tomrndom:
ref: https://tipit.avaza.com/task#!sortby=DateUpdatedDesc&task=3575965


smarcet (approved):
LGTM

PR #164, opened 2025-03-19
Hotfix/extra questions errors

gcutrini:
Previously, the onError callback was receiving a raw DOM element as its second argument rather than a proper React ref. This approach was problematic because a direct DOM element may not persist reliably between renders, which can lead to issues (for example, when trying to scroll to or focus on the error element).
This PR fixes that by providing an actual reference as onError second argument.

Also, this PR [fixes](https://github.com/OpenStackweb/openstack-uicore-foundation/commit/fa01f4f84b36eaa2515df18107b43594fb95493d) issue when Error was not being rendered after failed form submission.


smarcet (approved):
LGTM

PR #165, opened 2025-04-29
feat: implement as primary locking mechanism web lock api

smarcet:
super lock as fallback

see https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API

ref https://tipit.avaza.com/project/view/367963#!tab=task-pane&task=3790252


romanetar (approved):
LGTM

PR #168, opened 2025-06-11
fix: adjust font size on company input v2 type

tomrndom:
ref: https://tipit.avaza.com/project/view/376250#!tab=task-pane&task=3805705


smarcet (approved):
LGM

PR #195, opened 2026-02-05
feat: add createExternalStore utility and ClockProvider

gcutrini:
ref: https://app.clickup.com/t/86b8dm4da
Add a generic external store factory (`createExternalStore`) using `useSyncExternalStore` that lets components subscribe to frequently-updating data sources without unnecessary re-renders.

Includes a pre-built clock store (`ClockProvider`, `useClock`, `useClockSelector`) wrapping the existing Clock component. This enables projects to move clock state out of Redux, eliminating per-second re-renders of all connected components.

### API

- `createExternalStore(name)` → `{ Provider, useValue, useSelector }`
- `useValue()` — re-renders on every emit
- `useSelector(compute, isEqual)` — re-renders only when the computed result changes

### Clock usage

```js
import { ClockProvider, useClock, useClockSelector } from 'openstack-uicore-foundation/lib/components/clock-context';

<ClockProvider timezone={summit.time_zone_id}>
  <App />
</ClockProvider>

const nowUtc = useClock(); // re-renders every second
const allowedTickets = useClockSelector(compute, isEqual); // re-renders only on change
```

ref: https://app.clickup.com/t/86b8dm4da

## Summary by CodeRabbit

* **New Features**
  * Clock context with Provider plus hooks for server-synced time and selector-based subscriptions.
  * External-store factory and re-exports for easy creation and consumption of shared stores.

* **Documentation**
  * React compatibility note explaining the shim for React 16/17 and guidance for React 18+ migration.

* **Tests**
  * Comprehensive test suites for clock context and external-store behaviors.

* **Chores**
  * Added runtime dependency to support React 16/17 compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/openstack-uicore-foundation/pull/195?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/external-store.js
+++ b/src/utils/external-store.js
@@ -1,0 +1,201 @@
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * createExternalStore - Factory for creating React-optimized external stores.
+ *
+ * Problem:
+ *   When a frequently-updating data source (clock, WebSocket, polling, etc.)
+ *   pushes its value into shared state (global store, lifted useState, or a
+ *   context value), every consuming component re-renders on every update,
+ *   even when they only care about a derived condition that rarely changes.
+ *
+ * Solution:
+ *   createExternalStore() returns a Provider and hooks that store the value
+ *   in a ref (no re-renders) and use useSyncExternalStore so components
+ *   can opt in to updates selectively:
+ *
+ *   - useValue() → re-renders on every update
+ *   - useSelector(compute, isEqual) → re-renders only when the computed result changes
+ *
+ *   Components that don't call either hook are never affected by updates.
+ *
+ * How it works:
+ *   1. The Provider stores the value in a ref (writing to a ref never triggers
+ *      a React re-render) and keeps a Set of listener callbacks.
+ *   2. When emit(value) is called, the ref is updated and all listeners are
+ *      notified. These listeners come from useSyncExternalStore.
+ *   3. useSyncExternalStore (React 18, shimmed for 16/17) calls getSnapshot()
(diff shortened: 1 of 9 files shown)


src/utils/external-store.js
```diff
@@ -0,0 +1,201 @@
...
+            const value = getSnapshot();
+
+            if (value === lastValueRef.current) {
+                return lastResultRef.current;
+            }
+
+            const newResult = compute(value);
+            lastValueRef.current = value;
+
+            if (lastResultRef.current !== null && isEqual(lastResultRef.current, newResult)) {
```
smarcet: `null` is used both as the "not yet initialized" sentinel for `lastResultRef` (line 169) and as a possible legitimate return value from `compute`. Once `compute` returns `null`, subsequent `null → null` transitions skip the `isEqual` call entirely — `lastResultRef.current !== null` is `false`, so the custom equality function is never consulted for those transitions.

In practice, `useSyncExternalStore`'s own `Object.is` deduplication prevents spurious re-renders in all current cases, but the contract of the hook is broken: a consumer who passes a custom `isEqual` will find it silently ignored whenever the computed result is `null`.

Prefer a `Symbol` sentinel to cleanly separate the two states:
```js
const UNSET = Symbol();
const lastResultRef = useRef(UNSET);
// …
if (lastResultRef.current !== UNSET && isEqual(lastResultRef.current, newResult)) {
```

src/utils/external-store.js
```diff
@@ -0,0 +1,201 @@
...
+                return lastResultRef.current;
+            }
+
+            lastResultRef.current = newResult;
+            return newResult;
+        }, [getSnapshot, compute, isEqual]);
```
smarcet: `isEqual` is listed as a `useCallback` dependency, so when consumers pass an inline equality function the closure is recreated on every parent render. React's `useSyncExternalStore` responds to a changed `getSnapshot` identity by calling it immediately to check for tearing — correctness is preserved (the fast-path `value === lastValueRef.current` returns the cached result), but the extra call happens on every parent render rather than only on emits.

`compute` already receives ref-based stabilization (lines 174–177). The same pattern applied to `isEqual` eliminates the overhead and is consistent with how the rest of the hook is written:
```js
const lastIsEqualRef = useRef(isEqual);
lastIsEqualRef.current = isEqual; // always up-to-date, no cache invalidation needed

const getComputedValue = useCallback(() => {
    // …
    if (lastResultRef.current !== null && lastIsEqualRef.current(lastResultRef.current, newResult)) {
    // …
}, [getSnapshot, compute]); // isEqual removed from deps
```

smarcet (changes requested):
@gcutrini please review comments
also please check merge conflicts

smarcet (approved):
LGTM

PR #216, opened 2026-04-10
fix(show-confirm-dialog): eliminate react-dom/client build error on R…

smarcet:
ref: https://app.clickup.com/t/86b94xa4t

## Summary by CodeRabbit

* **New Features**
  * Added a global confirm-dialog component to manage and display confirmation dialogs consistently.

* **Behavior**
  * Dialog lifecycle now delegates to the global component, improving promise resolution and sequential dialog handling.

* **Tests**
  * Added comprehensive tests covering dialog display, confirm/cancel resolution, custom labels/colors, and multiple sequential dialogs.

* **Chores**
  * Updated version to 5.0.8-beta.3.


santipalenque (commented):
@smarcet this works but I think we overthinking this solution, there is no reason why we can just put the GlobalDialog component in <App /> so it renders always, and trigger the show/hide with a context solution. I think there is no good reason to spawn the dialog component and kill it every time we use it

src/components/mui/showConfirmDialog.js
```diff
@@ -11,27 +11,49 @@
...
-    // so consuming projects on React 16/17 won't get a "Module not found" error.
-    const mod = await import(/* webpackIgnore: true */ "react-dom/client");
-    createRootFn = mod.createRoot || null;
-  } catch (_) {
-    createRootFn = null;
-  }
-  return createRootFn;
-}
+// Module-level bridge: holds the callback registered by GlobalConfirmDialog
+let bridgeFn = null;
```
santipalenque: I guess this works, but I think the standard way in react would be to have a hook with a ref so that it keeps the value and can be used on client and in Dialog

santipalenque (approved):

tomrndom (approved):

PR #224, opened 2026-04-21
Feature/mui components improvements

santipalenque:
https://app.clickup.com/t/86b8k7d1g

## Summary by CodeRabbit

* **New Features**
  * New select (ComboBox) component (v2) with Formik integration
  * Several components now exposed as named exports

* **Improvements**
  * Dropdowns and timepicker accept explicit labels and placeholders
  * Standardized small sizing across form inputs
  * Improved empty-value placeholder display in selects
  * Better spacing in item modal and wider table cells for readability

* **Tests**
  * Added test mocks for select components

* **Chores**
  * Published beta release update


src/components/mui/formik-inputs/mui-formik-select-v2.js
```diff
@@ -0,0 +1,58 @@
...
+        name={name}
+        label={label}
+        labelId={`${name}-label`}
+        displayEmpty
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {...field}
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {...rest}
+        renderValue={(selected) => {
+          if (!selected) {
```
smarcet: @santipalenque  breaks for option value 0:
 The component is used in ItemTableField.js for the "ComboBox" case:
```JS
  options={field.values.map((v) => ({ value: v.id, label: v.value }))}
```
 v.id comes from server-side form field values. If any field value has id: 0, the user selects it, Formik stores 0, then renderValue receives 0, evaluates !0 as true, and renders the placeholder
  instead of the option label. The selection is silently swallowed visually , the value IS stored in Formik state, but the user sees a placeholder as if nothing was selected.

The fix is to replace the falsy check with an explicit check for "no meaningful selection":
```JS
  if (selected == null || selected === "") {
````
  This correctly treats 0, false, or any other legitimate option value as a real selection, while still showing the placeholder for null, undefined, and empty string.

smarcet (changes requested):
@santipalenque please review

santipalenque:
done @smarcet

smarcet (approved):
LGTM

PR #231, opened 2026-04-30
feat: add new mui formik color picker component

tomrndom:
ref: https://app.clickup.com/t/86b9n7p5w
* fix: add prop to fetch only initial options on async-select

image
image

## Summary by CodeRabbit

* **New Features**
  * Async select: optional local filtering for fast, case-insensitive client-side matching (reduces remote lookups).
  * Async select: improved multi-select selection and display behavior for more consistent formatting.
  * Formik color input: new color picker with local edit state, debounced commits, commit-on-blur, clear control, validation display, and color swatch.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/openstack-uicore-foundation/pull/231?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/mui/formik-inputs/mui-formik-async-select.js
+++ b/src/components/mui/formik-inputs/mui-formik-async-select.js
@@ -31,7 +31,8 @@ const MuiFormikAsyncAutocomplete = ({
   formatOption = (item) => ({ value: item.id.toString(), label: item.name }),
   formatSelectedValue = null,
   queryParams = [],
-  isMulti = false
+  isMulti = false,
+  defaultOptions
 }) => {
   const [field, meta, helpers] = useField(name);
   const [options, setOptions] = useState([]);
@@ -58,7 +59,7 @@ const MuiFormikAsyncAutocomplete = ({
   };
 
   useEffect(() => {
-    if (searchTerm) {
+    if (!defaultOptions && searchTerm) {
       const delayDebounce = setTimeout(() => {
         fetchOptions(searchTerm);
       }, DEBOUNCE_WAIT_250);
@@ -99,7 +100,16 @@ const MuiFormikAsyncAutocomplete = ({
       fullWidth
       getOptionLabel={(option) => option.label || ""}
       isOptionEqualToValue={(option, value) => option.value === value.value}
-      onInputChange={(e, newInput) => setSearchTerm(newInput)}
+      onInputChange={!defaultOptions ? (e, newInput) => setSearchTerm(newInput) : undefined}
+      filterOptions={
+        // only apply filterOptions for "local" search
+        defaultOptions
+          ? (options, { inputValue }) =>
+            options.filter((opt) =>
+              opt.label.toLowerCase().includes(inputValue.toLowerCase())
+            )
+          : undefined
+      }
       renderInput={(params) => (
         <TextField
           {...params}
--- a/src/components/mui/formik-inputs/mui-formik-async-select.js
+++ b/src/components/mui/formik-inputs/mui-formik-async-select.js
@@ -107,7 +107,7 @@ const MuiFormikAsyncAutocomplete = ({
       fullWidth
       getOptionLabel={(option) => option.label || ""}
       isOptionEqualToValue={(option, value) => option.value === value.value}
-      onInputChange={!localFilter ? (e, newInput) => setSearchTerm(newInput) : undefined}
+      onInputChange={!localFilter ? (e, newInput) => setSearchTerm(newInput) : (x) => x}
       filterOptions={
         // only apply filterOptions for "local" search
         localFilter
--- a/src/components/mui/formik-inputs/mui-formik-async-select.js
+++ b/src/components/mui/formik-inputs/mui-formik-async-select.js
@@ -20,33 +20,25 @@ import {
 } from "@mui/material";
 import { useField } from "formik";
 import { DEBOUNCE_WAIT_250 } from "../../../utils/constants";
-import PropTypes from "prop-types";
 
-/**
- * Async Autocomplete with two modes:
- * - Remote (default): fetches options from API on each user input (debounced).
- * - Local (localFilter=true): fetches once on mount and filters options client-side.
- * Note: localFilter mode assumes stable queryParams (set once on mount).
- * If queryParams need to change, remount the component instead.
- */
 const MuiFormikAsyncAutocomplete = ({
   name,
   queryFunction,
+  multiple = false,
   placeholder = "Select...",
   plainValue = false,
   hiddenOptions = [],
   formatOption = (item) => ({ value: item.id.toString(), label: item.name }),
   formatSelectedValue = null,
   queryParams = [],
-  isMulti = false,
-  localFilter = false
+  isMulti = false
 }) => {
   const [field, meta, helpers] = useField(name);
   const [options, setOptions] = useState([]);
   const [loading, setLoading] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
 
-  const value = field.value || (isMulti ? [] : null);
+  const value = field.value || (multiple ? [] : null);
   const error = meta.touched && meta.error;
 
   const fetchOptions = async (input = "") => {
@@ -66,7 +58,7 @@ const MuiFormikAsyncAutocomplete = ({
   };
 
   useEffect(() => {
-    if (!localFilter && searchTerm) {
+    if (searchTerm) {
       const delayDebounce = setTimeout(() => {
         fetchOptions(searchTerm);
       }, DEBOUNCE_WAIT_250);
@@ -80,7 +72,7 @@ const MuiFormikAsyncAutocomplete = ({
   }, []);
 
   const handleChange = (event, selected) => {
-    if (!isMulti) {
+    if (!multiple) {
       const selectedValue = plainValue ? selected?.value || "" : selected;
       helpers.setValue(selectedValue);
       return;
@@ -89,10 +81,10 @@ const MuiFormikAsyncAutocomplete = ({
     const selectedItems = plainValue
       ? selected.map((s) => s.value)
       : selected.map((s) =>
-        formatSelectedValue
-          ? formatSelectedValue(s)
-          : { id: parseInt(s.value), name: s.label }
-      );
+          formatSelectedValue
+            ? formatSelectedValue(s)
+            : { id: parseInt(s.value), name: s.label }
+        );
 
     helpers.setValue(selectedItems);
   };
@@ -107,18 +99,7 @@ const MuiFormikAsyncAutocomplete = ({
       fullWidth
       getOptionLabel={(option) => option.label || ""}
       isOptionEqualToValue={(option, value) => option.value === value.value}
-      onInputChange={!localFilter ? (e, newInput) => setSearchTerm(newInput) : (x) => x}
-      filterOptions={
-        // only apply filterOptions for "local" search
-        localFilter
-          ? (options, { inputValue }) =>
-            options.filter((opt) =>
-              String(opt.label ?? "").toLowerCase().includes(
-                String(inputValue ?? "").toLowerCase()
-              )
-            )
-          : undefined
-      }
+      onInputChange={(e, newInput) => setSearchTerm(newInput)}
       renderInput={(params) => (
         <TextField
           {...params}
@@ -148,21 +129,12 @@ const MuiFormikAsyncAutocomplete = ({
       )}
       renderOption={(props, option, { selected }) => (
         <li {...props}>
-          {isMulti && <Checkbox checked={selected} sx={{ mr: 1 }} />}
+          {multiple && <Checkbox checked={selected} sx={{ mr: 1 }} />}
           {option.label}
         </li>
       )}
     />
   );
-};
-
-MuiFormikAsyncAutocomplete.propTypes = {
-  name: PropTypes.string.isRequired,
-  isMulti: PropTypes.bool,
-  queryFunction: PropTypes.func.isRequired,
-  formatOption: PropTypes.func,
-  queryParams: PropTypes.array,
-  localFilter: PropTypes.bool,
 };
 
 export default MuiFormikAsyncAutocomplete;
--- a/src/components/mui/formik-inputs/mui-formik-async-select.js
+++ b/src/components/mui/formik-inputs/mui-formik-async-select.js
@@ -32,7 +32,6 @@ import PropTypes from "prop-types";
 const MuiFormikAsyncAutocomplete = ({
   name,
   queryFunction,
-  multiple = false,
   placeholder = "Select...",
   plainValue = false,
   hiddenOptions = [],
@@ -47,7 +46,7 @@ const MuiFormikAsyncAutocomplete = ({
   const [loading, setLoading] = useState(false);
   const [searchTerm, setSearchTerm] = useState("");
 
-  const value = field.value || (multiple ? [] : null);
+  const value = field.value || (isMulti ? [] : null);
   const error = meta.touched && meta.error;
 
   const fetchOptions = async (input = "") => {
@@ -81,7 +80,7 @@ const MuiFormikAsyncAutocomplete = ({
   }, []);
 
   const handleChange = (event, selected) => {
-    if (!multiple) {
+    if (!isMulti) {
       const selectedValue = plainValue ? selected?.value || "" : selected;
       helpers.setValue(selectedValue);
       return;
@@ -90,10 +89,10 @@ const MuiFormikAsyncAutocomplete = ({
     const selectedItems = plainValue
       ? selected.map((s) => s.value)
       : selected.map((s) =>
-          formatSelectedValue
-            ? formatSelectedValue(s)
-            : { id: parseInt(s.value), name: s.label }
-        );
+        formatSelectedValue
+          ? formatSelectedValue(s)
+          : { id: parseInt(s.value), name: s.label }
+      );
 
     helpers.setValue(selectedItems);
   };
@@ -114,7 +113,9 @@ const MuiFormikAsyncAutocomplete = ({
         localFilter
           ? (options, { inputValue }) =>
             options.filter((opt) =>
-              opt.label.toLowerCase().includes(inputValue.toLowerCase())
+              String(opt.label ?? "").toLowerCase().includes(
+                String(inputValue ?? "").toLowerCase()
+              )
             )
           : undefined
       }
@@ -147,7 +148,7 @@ const MuiFormikAsyncAutocomplete = ({
       )}
       renderOption={(props, option, { selected }) => (
         <li {...props}>
-          {multiple && <Checkbox checked={selected} sx={{ mr: 1 }} />}
+          {isMulti && <Checkbox checked={selected} sx={{ mr: 1 }} />}
           {option.label}
         </li>
       )}
@@ -157,6 +158,7 @@ const MuiFormikAsyncAutocomplete = ({
 
 MuiFormikAsyncAutocomplete.propTypes = {
   name: PropTypes.string.isRequired,
+  isMulti: PropTypes.bool,
   queryFunction: PropTypes.func.isRequired,
   formatOption: PropTypes.func,
   queryParams: PropTypes.array,
--- a/src/components/mui/formik-inputs/mui-formik-async-select.js
+++ b/src/components/mui/formik-inputs/mui-formik-async-select.js
@@ -20,7 +20,15 @@ import {
 } from "@mui/material";
 import { useField } from "formik";
 import { DEBOUNCE_WAIT_250 } from "../../../utils/constants";
+import PropTypes from "prop-types";
 
+/**
+ * Async Autocomplete with two modes:
+ * - Remote (default): fetches options from API on each user input (debounced).
+ * - Local (localFilter=true): fetches once on mount and filters options client-side.
+ * Note: localFilter mode assumes stable queryParams (set once on mount).
+ * If queryParams need to change, remount the component instead.
+ */
 const MuiFormikAsyncAutocomplete = ({
   name,
   queryFunction,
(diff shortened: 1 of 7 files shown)


src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,46 @@
...
+const MuiFormikColorInput = ({ name, ...rest }) => {
+  const [field, meta, helpers] = useField(name);
+  const [localValue, setLocalValue] = useState(field.value || "#000000");
+
+  useEffect(() => {
+    setLocalValue(field.value || "#000000");
+  }, [field.value]);
+
+  const handleChange = (e) => {
+    setLocalValue(e.target.value);
```
smarcet: @tomrndom 
MuiFormikColorInput desyncs from Formik between change and blur  
onChange only updates localValue; Formik's value updates only on blur. Native <input type="color"> typically commits via the picker without a blur (clicking outside the swatch does not always fire blur in all browsers/MUI variants). Submitting a form by pressing Enter/clicking submit while the picker is still open leaves Formik holding the old value. Validation also  won't run during interaction.

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -31,7 +31,8 @@ const MuiFormikAsyncAutocomplete = ({
   formatOption = (item) => ({ value: item.id.toString(), label: item.name }),
   formatSelectedValue = null,
   queryParams = [],
-  isMulti = false
+  isMulti = false,
+  defaultOptions
```
smarcet: @tomrndom 
Name collides with a well-known API and doesn't match its behavior 
 defaultOptions is the exact prop name used by react-select / common Autocomplete patterns, where it accepts an array of pre-loaded options (or true to preload). Here it's a plain boolean flag   
  for "local-only filtering." Consumers will reasonably assume they pass an array, and the implementation won't honor it — options state still comes from fetchOptions("") at line 71, not from the 
  prop.                                                                                                                                                                                             
                                                                                                                                                                                                    
  Suggested rename: localFilter (boolean) or useLocalFiltering.

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -58,7 +59,7 @@ const MuiFormikAsyncAutocomplete = ({
   };
 
   useEffect(() => {
```
smarcet, line 60: @tomrndom 
Stale options when queryParams change in local mode 
 useEffect(() => fetchOptions(""), []) at lines 70-72 runs once. The search-term effect at lines 60-67 is gated by !defaultOptions && searchTerm, so if queryParams (line 33) change after mount,  
  the list never refreshes. Either re-run the preload on queryParams change, or document that local mode requires stable params and a remount.

src/components/mui/formik-inputs/mui-formik-async-select.js
smarcet: **No PropTypes / no default value → silent behavior fork**

`defaultOptions` is destructured here without a default value, and the file has no `propTypes` block. A few concrete risks:

- Typos like `defaultoptions` silently keep remote behavior with no dev-mode warning.
- `[]` is truthy in JS, so `defaultOptions={[]}` flips the component to local-only mode with zero items — the opposite of what the prop name suggests.
- The same component now has two incompatible contracts hidden behind one boolean: with truthy `defaultOptions`, `queryFunction` runs only once on mount with `""`, `queryParams` changes are ignored, and `filterOptions` matches only on `option.label`.

Suggested:

```js
// in destructure
localFilter = false,
```

```js
// add at end of file
MuiFormikAsyncAutocomplete.propTypes = {
  name: PropTypes.string.isRequired,
  queryFunction: PropTypes.func.isRequired,
  localFilter: PropTypes.bool,
  // ...
};
```

Plus a one-line JSDoc above the component describing the two modes.

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  const handleBlur = (e) => {
+    helpers.setValue(e.target.value);
```
smarcet: **Default `#000000` masks "unset" and writes a value the user never picked**

When `field.value` is undefined, the swatch displays solid black (line 8) — indistinguishable from a user actively choosing black. This `handleBlur` then writes `#000000` into Formik even if the user never interacted with the picker, which:

- Defeats `required` validation (the field appears filled).
- Silently includes an unintended color in PATCH/POST payloads.

Suggested: only call `helpers.setValue` on actual user interaction — e.g. set a `dirtyRef` in `handleChange` and gate the blur write on it — or expose a `defaultValue` prop and treat blur-without-change as a no-op.

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,46 @@
...
+    helpers.setTouched(true);
+  };
+
+  return (
+    <TextField
+      type="color"
+      value={localValue}
+      onChange={handleChange}
+      onBlur={handleBlur}
```
smarcet: **Bypasses Formik's `field.onBlur` and drops `name` on the rendered input**

`useField(name)` returns `field` with the proper `name` and `onBlur`. This component sets explicit `value` / `onChange` / `onBlur` and never forwards `field` (or `name`) to the input. As a result:

- Formik's `field.onBlur` never fires — any cross-field validation hooks Formik wires up on blur are skipped.
- The rendered `<input>` has no `name` attribute, breaking browser autofill and any form-serialization fallback.

Suggested:

```jsx
<TextField
  type="color"
  name={name}
  {...field}
  onChange={handleChange}
  onBlur={(e) => { field.onBlur(e); helpers.setTouched(true); }}
  error={meta.touched && Boolean(meta.error)}
  helperText={meta.touched && meta.error}
  fullWidth
  sx={{ "& input[type='color']::-webkit-color-swatch-wrapper": { padding: "2px" } }}
  {...rest}
/>
```

smarcet (changes requested):
@tomrndom please review comments

src/components/mui/formik-inputs/mui-formik-color-input.js
tomrndom: The use of localValue is to improve component performance.

I could use a debounce on the handleChange to sync formik value with `localValue`

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,56 @@
+import React, { useRef, useState } from "react";
+import PropTypes from "prop-types";
+import { TextField } from "@mui/material";
+import { useField } from "formik";
+import { DEBOUNCE_WAIT_150 } from "../../../utils/constants";
+
+const MuiFormikColorInput = ({ name, ...rest }) => {
+  const [field, meta, helpers] = useField(name);
+  const [localValue, setLocalValue] = useState(field.value || "#000000");
```
smarcet, line 12: **[HIGH] Default `"#000000"` masks falsy `initialValues`.**

`useState(field.value || "#000000")` silently shows black while Formik state stays `""` / `null`. If the user never touches the input, the form submits with the original empty value while the swatch implies `#000000` was selected → data mismatch on save.

Suggested fix: also seed Formik on mount when `field.value` is empty:
```js
useEffect(() => { if (!field.value) helpers.setValue("#000000"); }, []);
```
or render the swatch only when `field.value` is set.

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,56 @@
...
+  const debounceRef = useRef(null);
```
smarcet, line 13: **[MEDIUM] No re-sync with external `field.value` changes.**

`localValue` is initialized once from `field.value` and only mutated by user input. After Formik `resetForm()`, parent `setFieldValue`, or async-loaded initial values, the swatch keeps showing the old color while Formik state has changed → broken edit/reset flows.

Suggested fix:
```js
useEffect(() => {
  if (field.value !== localValue) setLocalValue(field.value || "#000000");
}, [field.value]);
```
(guard against overwriting in-flight typing if needed.)

**[LOW] No debounce cleanup on unmount.**

A pending `setTimeout` can fire `helpers.setValue` after unmount → stale writes / dev warnings, especially in tabbed forms. Add:
```js
useEffect(() => () => { if (debounceRef.current) clearTimeout(debounceRef.current); }, []);
```

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -58,7 +67,7 @@ const MuiFormikAsyncAutocomplete = ({
...
-    if (searchTerm) {
+    if (!localFilter && searchTerm) {
```
smarcet: **[HIGH] `localFilter` mode filters against whatever `queryFunction("")` returns — usually a paged subset.**

The new docblock promises "fetches once on mount" + client-side filtering, but the only mount-time call (`fetchOptions("")` a few lines below) uses a `queryFunction` that elsewhere in this codebase honors server-side pagination/filter on `input`. Consumers will silently see an incomplete option list with no warning, contradicting the user-visible promise of "filter all options locally."

Suggested fix: in `localFilter` mode, require an explicit `fetchAll`/`pageSize` prop (or accept a pre-resolved `options` prop) and document the contract; otherwise emit a runtime warning.

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -99,7 +108,16 @@ const MuiFormikAsyncAutocomplete = ({
...
       getOptionLabel={(option) => option.label || ""}
       isOptionEqualToValue={(option, value) => option.value === value.value}
-      onInputChange={(e, newInput) => setSearchTerm(newInput)}
+      onInputChange={!localFilter ? (e, newInput) => setSearchTerm(newInput) : undefined}
+      filterOptions={
+        // only apply filterOptions for "local" search
+        localFilter
+          ? (options, { inputValue }) =>
+            options.filter((opt) =>
+              opt.label.toLowerCase().includes(inputValue.toLowerCase())
```
smarcet: **[MEDIUM] `opt.label.toLowerCase()` / `inputValue.toLowerCase()` are unguarded.**

The rest of the file already defends against missing labels (`option.label || ""` in `getOptionLabel`). One bad row from the API (`label` null/non-string) and the whole dropdown crashes. CodeRabbit flagged the same thing; not addressed.

Suggested fix:
```js
(options, { inputValue }) =>
  options.filter((opt) =>
    String(opt.label ?? "").toLowerCase().includes(
      String(inputValue ?? "").toLowerCase()
    )
  )
```

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -31,7 +39,8 @@ const MuiFormikAsyncAutocomplete = ({
   formatOption = (item) => ({ value: item.id.toString(), label: item.name }),
   formatSelectedValue = null,
   queryParams = [],
-  isMulti = false
+  isMulti = false,
```
smarcet: **[MEDIUM] Pre-existing `multiple` vs `isMulti` dual-prop divergence (worth fixing while editing this file).**

- `multiple` (default `false`, line 35) drives the `handleChange` value-shape branch (single vs array).
- `multiple={isMulti}` (further down) is what's passed to MUI `Autocomplete`.

A consumer passing only `isMulti={true}` (without `multiple={true}`) gets multi-select UI with **single-value Formik writes** → array/object collisions on submit. The new `propTypes` block cements this two-name API into the public surface.

Suggested fix: collapse to one prop (`multiple`), or alias one to the other in destructuring (`multiple = isMulti`), and update propTypes accordingly.

smarcet (commented):
Deep review — 7 findings (2 HIGH, 3 MEDIUM, 2 LOW). Inline comments on the relevant lines below. One non-inline note: no tests are added for either new behavior — `localFilter` (gated effect + filterOptions branch) and the color input (debounce + blur-flush + touched-on-blur) are stateful enough to regress silently. Project already has Jest + RTL set up; suggest one minimal RTL test per component.

smarcet (changes requested):
@tomrndom please review new comments

src/components/mui/formik-inputs/mui-formik-async-select.js
tomrndom: This can be set on the implementation as part of the queryParams according to the queryFunction used. 

Since the order of the per_page parameter is not always in the same position is a bit difficult to use a default MAX_PAGE_SIZE for all queryActions

src/components/mui/formik-inputs/mui-formik-color-input.js
tomrndom: The color picker would pick a default color if the user don't interact with the component. It looks like this contradicts https://github.com/OpenStackweb/openstack-uicore-foundation/pull/231#discussion_r3182654850.
Please let me know if I should add some initial state to display the input as not selected since the input type="color" doesn't support an empty value

tomrndom: PR updated with changes to display a placeholder if not value is selected and allow to clean value.

image
image

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,111 @@
...
+const MuiFormikColorInput = ({ name, placeholder = "Select a color", ...rest }) => {
+  const [field, meta, helpers] = useField(name);
+  const [hasValue, setHasValue] = useState(Boolean(field.value));
+  const [localValue, setLocalValue] = useState(field.value || "#000000");
+  const debounceRef = useRef(null);
+
+  useEffect(() => {
+    setHasValue(Boolean(field.value));
+    if (field.value && field.value !== localValue) setLocalValue(field.value);
+  }, [field.value]);
```
smarcet: @tomrndom here localValue is missing dep

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,111 @@
...
+              <IconButton size="small" onClick={handleClear} edge="end" disableRipple>
+                <ClearIcon fontSize="small" />
+              </IconButton>
+            </InputAdornment>
+          ),
+        } : undefined}
+        sx={{
+          "& input[type='color']::-webkit-color-swatch-wrapper": { padding: "2px" },
+        }}
+        {...rest}
```
smarcet, line 81: @tomrndom 
```JS                                                                                                                        
  InputProps={hasValue ? { endAdornment: <ClearIcon .../> } : undefined}                                                                                                                            
  sx={{ ... }}                                                                                                                                                                                      
  {...rest} 
  // consumer's InputProps wins, 
```
clear button disappears                                                                                                                                
 MUI does not merge InputProps , last value wins. A consumer passing InputProps (e.g. a start adornment) silently drops the clear button. Fix: 
```JS
destructure InputProps as restInputProps from rest  
```
and merge:      
```JS                                                                                                                                                                                  
  InputProps={{ ...(hasValue ? { endAdornment: ... } : {}), ...restInputProps }}         
```

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -99,7 +107,18 @@ const MuiFormikAsyncAutocomplete = ({
       fullWidth
       getOptionLabel={(option) => option.label || ""}
       isOptionEqualToValue={(option, value) => option.value === value.value}
-      onInputChange={(e, newInput) => setSearchTerm(newInput)}
+      onInputChange={!localFilter ? (e, newInput) => setSearchTerm(newInput) : undefined}
+      filterOptions={
```
smarcet: @tomrndom 
replace undefined with (x) => x. 
passing (x) => x tells MUI "these options are already filtered, show them all." This is the standard pattern recommended in the MUI docs for server-side autocomplete.

smarcet (changes requested):
@tomrndom please review

smarcet (changes requested):
@tomrndom please review comments

src/components/mui/formik-inputs/mui-formik-async-select.js
```diff
@@ -20,25 +20,33 @@ import {
 } from "@mui/material";
 import { useField } from "formik";
 import { DEBOUNCE_WAIT_250 } from "../../../utils/constants";
+import PropTypes from "prop-types";
 
+/**
+ * Async Autocomplete with two modes:
+ * - Remote (default): fetches options from API on each user input (debounced).
+ * - Local (localFilter=true): fetches once on mount and filters options client-side.
```
santipalenque: Local mode makes this autocomplete NOT ASYNC . I strongly suggest to create a new component when the functionality changes drastically like in this case.

santipalenque: Also this change seems out of scope, what is this used for ?

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,113 @@
+import React, { useEffect, useRef, useState } from "react";
+import PropTypes from "prop-types";
+import { Box, IconButton, InputAdornment, TextField } from "@mui/material";
+import ClearIcon from "@mui/icons-material/Clear";
+import { useField } from "formik";
+import { DEBOUNCE_WAIT_150 } from "../../../utils/constants";
+
+const MuiFormikColorInput = ({ name, placeholder = "Select a color", InputProps: restInputProps, ...rest }) => {
```
santipalenque: use translations for literals

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,113 @@
...
+  }, [field.value, localValue]);
+
+  useEffect(() => () => {
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+  }, []);
+
+  const handleChange = (e) => {
+    const value = e.target.value;
+    setLocalValue(value);
+    setHasValue(true);
```
santipalenque, line 27: why do we need this state ? can't you just use localValue ?

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,113 @@
...
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+    debounceRef.current = setTimeout(() => {
```
santipalenque, line 29: why do we need to debounce the onchange ? we never do this . If this is strictly necessary, add comments why

src/utils/constants.js
```diff
@@ -6,6 +6,7 @@ export const ZERO_INT = 0;
 
 export const CODE_200 = 200;
 
+export const DEBOUNCE_WAIT_150 = 150;
 export const DEBOUNCE_WAIT_250 = 250;
```
santipalenque, line 11: any reason why you can't just use 250 ?

santipalenque (changes requested):
a few questions and suggestions. Also looks like you are missing the tests for the color input

src/components/mui/formik-inputs/mui-formik-async-select.js
tomrndom: I'll create a new PR with the changes for the autocomplete. It started with a new prop to filter locally since this is required by the new company popup on summit admin (input for countries), but it ended requiring so much changes that is gonna be better to treat it in a separate PR.

src/components/mui/formik-inputs/mui-formik-color-input.js
tomrndom: This is to limit the render of formik props when the user is dragging the cursor to pick a color, without this the component turns laggy.

src/components/mui/formik-inputs/mui-formik-color-input.js
tomrndom: this state is to track when the user pick a color since html color input requires a default value

src/utils/constants.js
tomrndom: 250 made the delay between selecting a color and seeing it applied slightly noticeable, 150 felt a bit more responsive, but both work fine

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,113 @@
...
+const MuiFormikColorInput = ({ name, placeholder = "Select a color", InputProps: restInputProps, ...rest }) => {
+  const [field, meta, helpers] = useField(name);
+  const [hasValue, setHasValue] = useState(Boolean(field.value));
+  const [localValue, setLocalValue] = useState(field.value || "#000000");
+  const debounceRef = useRef(null);
+
+  useEffect(() => {
+    setHasValue(Boolean(field.value));
+    if (field.value && field.value !== localValue) setLocalValue(field.value);
+  }, [field.value, localValue]);
```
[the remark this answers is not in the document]
santipalenque: please check if this is valid @tomrndom

src/components/mui/formik-inputs/mui-formik-color-input.js
santipalenque: ok cool, then put a comment on the code please. Also this means that if I move the cursor slowly it will change the color multiple times ? I think we should try to find a better solution than debouncing

src/components/mui/formik-inputs/mui-formik-color-input.js
```diff
@@ -0,0 +1,114 @@
...
+            ),
+          } : {}), ...restInputProps
+        }}
+        sx={{
+          "& input[type='color']::-webkit-color-swatch-wrapper": { padding: "2px" },
+        }}
+        {...rest}
+      />
+      {!hasValue && (
+        <Box
```
santipalenque, line 84: is this covering the color input when no value is selected ? There must be a better way, this is pretty dirty

santipalenque (commented):
@smarcet @tomrndom I think this solution is too hacky and doesn't seem robust. I wonder why we not just use something like [this](https://www.npmjs.com/package/mui-color-input).  Also @tomrndom is there a branch where I can test this ?

smarcet:
> @smarcet @tomrndom I think this solution is too hacky and doesn't seem robust. I wonder why we not just use something like [this](https://www.npmjs.com/package/mui-color-input). Also @tomrndom is there a branch where I can test this ?

yes i do agree lets close this one and use that lib
related show admin PR is https://github.com/fntechgit/summit-admin/pull/910

PR #241, opened 2026-05-15
fix(company-input-v2): commit typed/autofilled value on blur

gcutrini:

## Summary

`CompanyInputV2` wraps MUI Autocomplete in `freeSolo` mode, but `onChange` only fires when the user explicitly selects an option. Values entered via typing-and-tabbing or browser autofill (notably **iOS Chrome**) never propagate to the parent, so required-field validation fails on submit.

Reproduce with autofill simulation in DevTools: set the input value via the native setter, dispatch `input`, then blur — without this PR the parent's `company` stays `{id: null, name: ""}`; with it the parent receives `{id: 0, name: <typed>}`.

## Changes

- **Add `autoSelect`** to the Autocomplete. MUI's intended mechanism for committing the current input value on blur in a free-text combo.
- **Normalize strings in `onChange`** to the `{id, name}` shape consumers expect, since `autoSelect` calls `onChange` with the raw typed/autofilled string.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved company input field to properly normalize manually typed values, including automatic whitespace trimming and consistent formatting when input changes occur.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/openstack-uicore-foundation/pull/241?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/inputs/company-input-v2.js
+++ b/src/components/inputs/company-input-v2.js
@@ -56,6 +56,7 @@ const CompanyInputV2 = ({ summitId, isRequired, sx, onChange, id, name, label, v
       name={name}
       options={options}
       autoComplete
+      autoSelect
       freeSolo
       includeInputInList
       filterSelectedOptions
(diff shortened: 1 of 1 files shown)


smarcet (approved):
LGTM

PR #246, opened 2026-05-21
fix(company-input-v2): dropdown UX cleanup, clear-icon, canonical-case fixes

gcutrini:

## Summary

Reworks the CompanyInputV2 field UX (after #241 enabled autoSelect on blur).

- 96abac9 show plain options, pick canonical case on blur. Removes the extra "no match / use typed value" row from the dropdown, and makes blur commit the canonical company when the typed text matches one regardless of case (e.g. typing "tipit" picks "Tipit").

- 243150c match canonical case post-API when blur beats the response. If the user types and blurs before the search request comes back, the typed text gets accepted as-is. Once the response arrives, if it included a matching company, the field swaps to that company automatically.

- e3759dc hide clear icon when value is effectively empty. Treats blank values (empty string, value object with no name) as no selection so MUI stops showing the clear (x) icon on hover of a field that looks empty.


src/components/inputs/__tests__/company-input-v2.test.js
```diff
@@ -0,0 +1,137 @@
+/**
+ * Copyright 2025 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+import {
```
smarcet: @gcutrini 
Missing coverage:                                                                                                                                                                               
    - Blur triggers canonical-case swap through the real component (the core PR feature , tested nowhere).                                                                                          
    - Post-API response swap when blur precedes the response (the race-condition fix , tested nowhere).                                                                                             
    - Clear icon hidden when value is an empty-name object (the third headline fix , tested nowhere).

src/components/inputs/company-input-v2.js
```diff
@@ -13,41 +13,82 @@
...
+    const normalizedValue = React.useMemo(() => normalizeCompanyValue(value), [value]);
 
     React.useEffect(() => {
         if (inputValue === "") {
-            setOptions(value ? [value] : []);
+            setOptions(normalizedValue ? [normalizedValue] : []);
             return undefined;
         }
 
         queryRegistrationCompanies(summitId, inputValue, (results) => {
```
smarcet, line 77: @gcutrini 
Stale-closure race: late API callback fires onChange after user clears the field
No cleanup in the effect; the in-flight queryRegistrationCompanies callback closes over the old normalizedValue ({id:0, name:"tipit"}), so if the user clears the field before the response arrives, the callback still fires isNewCompany(normalizedValue) as true and calls onChange with the canonical company , overwriting the user's deliberate clear action.

suggested fix
Add a cancellation flag and return a cleanup function: 

```JS
 React.useEffect(() => {                                                                                                                                                                           
      if (inputValue === "") { ... return undefined; }                                                                                                                                              
                                                                                                                                                                                                    
      let cancelled = false; // ← flag                                                                                                                                                              
                                                                                                                                                                                                    
      queryRegistrationCompanies(summitId, inputValue, (results) => {                                                                                                                               
          if (cancelled) return; // ← bail if effect was re-run or component unmounted                                                                                                              
                                                                                                                                                                                                    
          ...                                                                                                                                                                                       
          if (isNewCompany(normalizedValue)) {                                                                                                                                                      
              const match = findExistingByName(results, normalizedValue.name);                                                                                                                      
              if (match) {                                                                                                                                                                          
                  onChange({ target: { id: name, value: match, ... } });                                                                                                                            
              }                                                                                                                                                                                     
          }                                                                                                                                                                                         
      }, options2Show);                                                                                                                                                                             
                                                                                                                                                                                                    
      return () => { cancelled = true; }; // ← cleanup: runs before next effect or on unmount                                                                                                       
  }, [normalizedValue, inputValue, summitId, options2Show, onChange, name]);
```

src/components/inputs/company-input-v2.js
```diff
@@ -13,41 +13,82 @@
...
+                if (match) {
+                    onChange({
+                        target: { id: name, value: match, type: "companyinput" }
+                    });
+                }
+            }
         }, options2Show);
         return undefined;
-    }, [value, inputValue, summitId, options2Show]);
+    }, [normalizedValue, inputValue, summitId, options2Show, onChange, name]);
```
smarcet: @gcutrini if u add the onChange on a deps array this will trigger an infinite loop on the effect
generally speaking most of the clients pass and inline arrow function and that arrow function object is a new one on each re render
so remove the onChange deps to avoid the infinite loop or use a a ref to hold the latest version of the onChange callback

smarcet (changes requested):
@gcutrini please review

gcutrini:
@smarcet — pushed the changes:

- Added a `useEventCallback` util and used it to wrap the parent's `onChange`, so the effect no longer re-runs on inline arrow identities; `onChange` and `name` are out of the deps array.
- Cancellation flag + cleanup in the effect so a late API response can no longer overwrite a user's clear/change.
- 3 integration tests through the real component: blur canonical-case swap, post-API replacement when blur beats the response, clear icon hidden for empty-name value.

Ready for another look.

src/components/inputs/company-input-v2.js
```diff
@@ -141,43 +161,15 @@ const CompanyInputV2 = ({ summitId, isRequired, sx, onChange, id, name, label, v
...
-                                sx={{ color: "text.secondary", padding: "5px 0" }}
-                            >
-                                {option.name}
-                            </Typography>
-                        )}
+                    <li key={key} {...optionProps}>
+                        <Typography
+                            variant="body2"
+                            sx={{ fontSize: "1em", color: "text.secondary", padding: "5px 0" }}
+                        >
```
smarcet, line 179: **[LOW] `renderOption` renders blank for string-valued options**

`option.name` is `undefined` when `option` is a string. `normalizeCompanyValue` preserves non-empty strings, so the `options` array can contain plain string entries (e.g. when the consumer passes `value="Tipit"`). MUI calls `renderOption` for those entries and `<Typography>` renders nothing.

`getOptionLabel` already handles this correctly — `renderOption` needs the same guard:

```jsx
renderOption={(props, option) => {
    const { key, ...optionProps } = props;
    const label = typeof option === "string" ? option : option.name;
    return (
        <li key={key} {...optionProps}>
            <Typography variant="body2" sx={{ fontSize: "1em", color: "text.secondary", padding: "5px 0" }}>
                {label}
            </Typography>
        </li>
    );
}}
```

src/components/inputs/company-input-v2.js
```diff
@@ -60,71 +114,37 @@ const CompanyInputV2 = ({ summitId, isRequired, sx, onChange, id, name, label, v
...
+                // autoSelect commits the raw typed/autofilled string on blur. If the
+                // string matches an existing company case-insensitively, pick that
+                // option (so typing "tipit" tabs out to "Tipit"). Otherwise commit
+                // the typed value as a free-text {id: 0, name} entry.
                 if (typeof tmpValue === "string" && tmpValue.trim()) {
-                    tmpValue = { id: 0, name: tmpValue.trim() };
+                    const trimmed = tmpValue.trim();
+                    tmpValue = findExistingByName(options, trimmed) || { id: 0, name: trimmed };
                 }
                 setOptions(tmpValue ? [tmpValue, ...options] : options);
```
smarcet: **[LOW] Blur-committed company is duplicated in the options list**

When `findExistingByName` resolves `tmpValue` to an existing company, that company is already present in `options` (it came from the prior API call). Prepending it unconditionally produces a duplicate entry the next time the dropdown opens.

```js
// before: may duplicate an existing entry
setOptions(tmpValue ? [tmpValue, ...options] : options);

// after: filter out the same id before prepending
setOptions(tmpValue
    ? [tmpValue, ...options.filter(o => o?.id !== tmpValue?.id)]
    : options
);
```

This is visual-only and resolves itself on the next `onInputChange`, but it is noticeable when the user re-opens the dropdown immediately after blur.

smarcet (changes requested):
@gcutrini please re review

smarcet (approved):
LGTM

PR #250, opened 2026-05-21
fix(company-input-v2): dropdown UX cleanup, clear-icon, canonical-case fixes (v5 port of #246)

gcutrini:
ref: https://app.clickup.com/t/86b9rnr4r

Port of #246 to main (v5). Functionally identical to the v4.x version, re-indented to match the v5 codebase style and reusing the @testing-library/react@12 that v5 already has installed (no new deps).

21 tests, all passing.

## Summary by CodeRabbit

* **New Features**
  * Added a stable event-callback hook for safer, identity-stable callbacks in components.

* **Bug Fixes**
  * Improved company input behavior: case-insensitive canonicalization, safer option rendering, de-duplication, and guarded async responses to avoid late-state surprises.

* **Tests**
  * Added comprehensive test suite covering company input validation, lookup/normalization helpers, and integration behaviors.

* **Chores**
  * Exposed the new hook to the build so it can be consumed by the app.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/openstack-uicore-foundation/pull/250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #253, opened 2026-05-22
test(company-input-v2): pull api callback from mock call args to avoid CI race

gcutrini:
Same test-stability fix that was applied to the v5 port (PR #250). The original `auto-replaces a free-text commit…` test from #246 captures the API callback via a closure variable that gets set inside the mock implementation; the closure is read on the next line, which races against React's effect commit timing. Passes locally, can flake in CI under different timing.

Pulling the callback directly from `mock.calls[last][2]` makes the test deterministic.


gcutrini:
Closing as out-of-scope. The flaky-closure pattern in this v4 test is real but not what Sebastian flagged; we can revisit if it actually flakes in CI.

PR #255, opened 2026-05-26
Fix: global qty field prevent negatives

santipalenque:
When clicking down arrow on a globalQuantityField ,the app crashes. This PR is to prevent down arrow click on a GlobalQuantityField when value is 0. 

## Summary by CodeRabbit

* **New Features**
  * Quantity inputs now validate user input, prevent negative entries, reset cleared fields to 0, and clamp values to configured per-item limits when present.

* **Bug Fixes**
  * Table column sizing for item rows and headers standardized for more consistent layout.

* **Tests**
  * Expanded test coverage for quantity behavior, clamping, and reset cases.

* **Chores**
  * Updated package version to 5.0.25-beta.1.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/openstack-uicore-foundation/pull/255?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/components/mui/FormItemTable/components/GlobalQuantityField.js
+++ b/src/components/mui/FormItemTable/components/GlobalQuantityField.js
@@ -36,11 +36,10 @@ const GlobalQuantityField = ({
 
   const handleChange = (e) => {
     const val = parseInt(e.target.value, 10);
+    if (isNaN(val)) { helpers.setValue(0); return; }
     const max = row.quantity_limit_per_sponsor;
-    if (!isNaN(val)) {
-      const clamped = max !== undefined ? Math.min(Math.max(val, 0), max) : Math.max(val, 0);
-      helpers.setValue(clamped);
-    }
+    const clamped = max !== undefined ? Math.min(Math.max(val, 0), max) : Math.max(val, 0);
+    helpers.setValue(clamped);
   };
 
   return (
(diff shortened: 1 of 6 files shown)


src/components/mui/FormItemTable/components/GlobalQuantityField.js
```diff
@@ -34,13 +34,23 @@ const GlobalQuantityField = ({
     helpers.setTouched(true);
   }, [value]);
 
+  const handleChange = (e) => {
```
smarcet, line 37: @santipalenque 
Missing tests:                                                                                                                                                                                  
    - Value clamped when user types above quantity_limit_per_sponsor                                                                                                                                
    - Value resets to 0 (not silent no-op) when field is cleared or down-arrow used from 0                                                                                                          
    - Behavior when quantity_limit_per_sponsor is 0 or undefined

src/components/mui/FormItemTable/components/GlobalQuantityField.js
```diff
@@ -34,13 +34,23 @@ const GlobalQuantityField = ({
...
+    const val = parseInt(e.target.value, 10);
```
smarcet, line 38: @santipalenque 
 parseInt("", 10) returns Nan
, so if (!isNaN(val)) silently drops the event ,Formik retains the last value while the DOM shows empty. On submit, the stale value is sent
fix
if (isNaN(val)) { helpers.setValue(0); return; } so clearing resets to the floor value

smarcet (changes requested):
@santipalenque please review comments

smarcet (approved):
LGTM

PR #261, opened 2026-06-02
fix: match inputs with figma styles

priscila-moneo:
ref: https://app.clickup.com/t/9014802374/86b8tt4df

image

## Summary by CodeRabbit

* **New Features**
  * Introduced enhanced Material-UI theme customization with improved border and icon styling across components.

* **Style**
  * Updated search input border colors for better visual consistency.
  * Refined select dropdown icon appearance and dimensions.


src/components/mui/formik-inputs/mui-formik-dropdown-checkbox.js
```diff
@@ -59,6 +60,7 @@ const MuiFormikDropdownCheckbox = ({ name, label, placeholder, options, ...rest
         label={label}
         labelId={`${name}-label`}
         multiple
+        IconComponent={KeyboardArrowDownIcon}
```
santipalenque: was this a request ? I cannot see this anywhere on the ticket or the figma

src/components/mui/formik-inputs/mui-formik-dropdown-radio.js
```diff
@@ -44,6 +45,7 @@ const MuiFormikDropdownRadio = ({ name, label, options, placeholder, ...rest })
         name={name}
         label={label}
         labelId={`${name}-label`}
+        IconComponent={KeyboardArrowDownIcon}
```
santipalenque: if we are going to change all dropdown please use CustomTheme:

`const theme = createTheme({
  components: {
    MuiSelect: {
      defaultProps: {
        // Change the dropdown arrow icon globally
        IconComponent: KeyboardArrowDownIcon, 
      },
    },
  },
});`

santipalenque (changes requested):
Please move all these changes to CustomTheme

src/components/mui/formik-inputs/mui-formik-dropdown-checkbox.js
priscila-moneo: Yes, it is.
Ticket Request:
image

Figma Screenshot:
image

src/utils/theme.js
```diff
@@ -0,0 +1,49 @@
...
+          fontSize: "24px",
+          width: "24px",
+          height: "24px"
+        }
+      }
+    }
+  }
+};
+
+export const CustomThemeBase = createTheme(uiCoreThemeOverrides);
```
santipalenque: I don't think we need to createTheme in uicore, just export the definitions and let the client app use it as they need. Also please name it MuiBaseCustomTheme

src/components/mui/search-input.js
```diff
@@ -83,7 +83,10 @@ const SearchInput = ({ term, onSearch, placeholder = "Search...", debounced }) =
         "& .MuiOutlinedInput-root": {
           height: "36px",
           paddingX: 1
-        },        
+        },
+        "& .MuiOutlinedInput-root .MuiOutlinedInput-notchedOutline": { borderColor: "rgba(0, 0, 0, 0.23)" },
```
santipalenque: @smarcet another example, we need to standarize these colors and use one in the color palette or use a custom one set on CustomTheme. If for some reason we don't want to do that because is too specific, then we should use one of the definitions gray[500] or something. In this particular case @priscila-moneo looks like this color is the default for border color on outline variant so I don't understand why are you overriding with that same color

santipalenque (changes requested):

src/components/mui/search-input.js
priscila-moneo: The override is intentional. The outlined variant default border color was changed to #000, but according to the current Figma designs, the Search input should keep using the previous border color instead of inheriting the new default. That's why the color is explicitly set here.
image
image

JpMaxMan: Consistency is key in design systems.  We should check with design team and see which one should be used and if it's differentiated by input type categorize that difference so it can be referenced the same way by all input types of that category.

santipalenque (approved):
Looks good @priscila-moneo , I'll pre-approve but can you move the MuiBaseCustomTheme to /components/mui/MuiBaseCustomTheme.js ? I don't think this should go in utils. Also remember to update the export

smarcet (approved):
LGTM

PR #263, opened 2026-06-05
feat: upload input v3 image preview improvement

priscila-moneo:
ref: https://app.clickup.com/t/86ba8wmgh

## Summary by CodeRabbit

* **New Features**
  * Immediate image previews for uploading files, including quick display as files are added.

* **Improvements**
  * Faster, more reliable preview rendering for local `data:`/`blob:` sources.
  * Safer handling of upload cancellation and errors by revoking unused previews and preventing preview carry-over.
  * Improved preview persistence across updates, supporting server-renamed uploads and out-of-order completions.

* **Tests**
  * Added/expanded coverage for upload preview lifecycle and progressive image loading/race-condition behavior.

### DIFF
--- a/src/components/inputs/upload-input-v3/index.js
+++ b/src/components/inputs/upload-input-v3/index.js
@@ -11,7 +11,7 @@
  * limitations under the License.
  **/
 
-import React, { useState, useRef, useMemo, useCallback, useEffect } from 'react';
+import React, { useState, useRef, useMemo, useCallback, useLayoutEffect } from 'react';
 import T from "i18n-react/dist/i18n-react";
 import {
   Box,
@@ -29,6 +29,19 @@ import { DropzoneV3 } from './dropzone-v3';
 import ProgressiveImg from '../../progressive-img';
 import file_icon from '../upload-input/file.png';
 import './index.less';
+
+const fileRowSx = {
+  display: 'flex',
+  alignItems: 'center',
+  py: 1.5,
+  mb: 1,
+};
+
+const formatFileSize = (bytes) => {
+  if (!bytes) return '0 KB';
+  if (bytes >= 1024 * 1024) return `${Math.round(bytes / (1024 * 1024))} MB`;
+  return `${Math.round(bytes / 1024)} KB`;
+};
 
 const UploadInputV3 = ({
   value = [],
@@ -55,26 +68,17 @@ const UploadInputV3 = ({
   const dropzoneInstanceRef = useRef(null);
   const [uploadingFiles, setUploadingFiles] = useState([]);
   const [errorFiles, setErrorFiles] = useState([]);
+  const [filePreviews, setFilePreviews] = useState({});
 
-  const getDefaultAllowedExtensions = useCallback(() => {
-    return mediaType && mediaType.type
-      ? mediaType?.type?.allowed_extensions.map((ext) => `.${ext.toLowerCase()}`).join(",")
-      : '';
-  }, [mediaType]);
+  const allowedExt = useMemo(() => {
+    if (getAllowedExtensions) return getAllowedExtensions();
+    return mediaType?.type?.allowed_extensions?.map(ext => `.${ext.toLowerCase()}`).join(',') ?? '';
+  }, [getAllowedExtensions, mediaType]);
 
-  const getDefaultMaxSize = useCallback(() => {
-    return mediaType ? mediaType?.max_size / (1024 * 1024) : 100;
-  }, [mediaType]);
-
-  const allowedExt = useMemo(() =>
-    getAllowedExtensions ? getAllowedExtensions() : getDefaultAllowedExtensions(),
-    [getAllowedExtensions, getDefaultAllowedExtensions]
-  );
-
-  const maxSize = useMemo(() =>
-    getMaxSize ? getMaxSize() : getDefaultMaxSize(),
-    [getMaxSize, getDefaultMaxSize]
-  );
+  const maxSize = useMemo(() => {
+    if (getMaxSize) return getMaxSize();
+    return mediaType ? mediaType.max_size / (1024 * 1024) : 100;
+  }, [getMaxSize, mediaType]);
 
   const canUpload = useMemo(() =>
     !maxFiles || value.length < maxFiles,
@@ -114,13 +118,7 @@ const UploadInputV3 = ({
     media_upload: value,
   }), [mediaType, value]);
 
-  const formatFileSize = useCallback((bytes) => {
-    if (!bytes) return '0 KB';
-    if (bytes >= 1024 * 1024) return `${Math.round(bytes / (1024 * 1024))} MB`;
-    return `${Math.round(bytes / 1024)} KB`;
-  }, []);
-
-  const formatExtensionsDisplay = useCallback(() => {
+  const extDisplay = useMemo(() => {
     if (!allowedExt) return '';
     const exts = allowedExt.split(',')
       .map(e => e.trim().replace('.', '').toUpperCase())
@@ -132,15 +130,21 @@ const UploadInputV3 = ({
 
   const handleRemove = useCallback((file) => (ev) => {
     ev.preventDefault();
+    const blobUrl = filePreviews[file.filename];
+    if (blobUrl) {
+      URL.revokeObjectURL(blobUrl);
+      setFilePreviews(prev => { const next = { ...prev }; delete next[file.filename]; return next; });
+    }
     onRemove(file);
-  }, [onRemove]);
+  }, [onRemove, filePreviews]);
 
   const handleDropzoneReady = useCallback((dz) => {
     dropzoneInstanceRef.current = dz;
   }, []);
 
   const handleAddedFile = useCallback((file) => {
-    setUploadingFiles(prev => [...prev, { name: file.name, size: file.size, progress: 0, complete: false }]);
+    const previewUrl = file.type?.startsWith('image/') ? URL.createObjectURL(file) : null;
+    setUploadingFiles(prev => [...prev, { name: file.name, size: file.size, progress: 0, complete: false, previewUrl }]);
     if (onUploadStart) onUploadStart(file);
   }, [onUploadStart]);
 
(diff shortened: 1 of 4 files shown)


src/components/progressive-img/index.js
```diff
@@ -26,23 +26,31 @@ import file_icon from "../inputs/upload-input/file.png";
  * @constructor
  */
 const ProgressiveImg = ({ placeholderSrc, src, ...props }) => {
-    const isCancelled = useRef(false);
```
santipalenque, line 29: are you sure to do this ? is not the same to use a Ref or a let

src/components/inputs/upload-input-v3/index.js
```diff
@@ -139,7 +142,14 @@ const UploadInputV3 = ({
...
 
   const handleAddedFile = useCallback((file) => {
-    setUploadingFiles(prev => [...prev, { name: file.name, size: file.size, progress: 0, complete: false }]);
+    setUploadingFiles(prev => [...prev, { name: file.name, size: file.size, progress: 0, complete: false, previewUrl: null }]);
+  }, []);
+
+  const handleThumbnail = useCallback((file, dataURL) => {
+    pendingPreviewsRef.current.push({ name: file.name, size: file.size, dataURL });
+    setUploadingFiles(prev => prev.map(f =>
+      f.name === file.name && f.size === file.size ? { ...f, previewUrl: dataURL } : f
```
santipalenque: what happens if changing for another file with same name and same size ?

src/components/inputs/upload-input-v3/index.js
```diff
@@ -54,6 +54,9 @@ const UploadInputV3 = ({
   const dropzoneInstanceRef = useRef(null);
   const [uploadingFiles, setUploadingFiles] = useState([]);
   const [errorFiles, setErrorFiles] = useState([]);
+  const [filePreviews, setFilePreviews] = useState({});
+  const prevValueRef = useRef(value);
+  const pendingPreviewsRef = useRef([]);
```
santipalenque: seems like filePreviews already has this information, why you need this ref ?

src/components/inputs/upload-input-v3/index.js
```diff
@@ -162,13 +172,39 @@ const UploadInputV3 = ({
     ));
   }, []);
 
-  // Once the parent updates value, remove all completed files from uploadingFiles
-  useEffect(() => {
+  // Once the parent updates value, remove completed uploading files and assign local previews to new files
+  useLayoutEffect(() => {
```
santipalenque, line 184: I don't see the use case for this change, can you explain ?

src/components/inputs/upload-input-v3/index.js
santipalenque: Simplification: collapse the preview state into a single store
  
  The current design stores each dataURL in two places simultaneously — uploadingFiles[i].previewUrl (for the in-progress row) and pendingPreviewsRef (so useLayoutEffect can re-key it by
  server filename later). This forces three separate cleanup sites that must all stay in sync: handleFileError, handleDeleteUploading, and useLayoutEffect. 
  All of this goes away if handleThumbnail stores directly into filePreviews keyed by the original filename:

  const handleThumbnail = useCallback((file, dataURL) => {
    setFilePreviews(prev => ({ ...prev, [file.name]: dataURL }));
  }, []);

  Then:
  - Uploading rows read filePreviews[file.name] at render time — previewUrl can be dropped from uploadingFiles and handleAddedFile no longer needs to set it.
  - Committed value rows already fall back to serverPreviewSrc, so the stem-match can be done inline at read time: filePreviews[filename] || filePreviews[Object.keys(filePreviews).find(k =>
  filename.includes(k.replace(/\.[^.]+$/, '')))] || serverPreviewSrc.
  - pendingPreviewsRef, prevValueRef, and the entire batched matching algorithm in useLayoutEffect are removed.
  - Cleanup on cancel/error becomes a single setFilePreviews delete instead of a coordinated mutation across two collections.

src/components/inputs/upload-input-v3/index.js
```diff
@@ -162,13 +172,39 @@ const UploadInputV3 = ({
...
+    const prevFilenames = new Set(prevValue.map(f => f.filename));
+    const newFiles = value.filter(f => !prevFilenames.has(f.filename));
+    if (newFiles.length > 0 && pendingPreviewsRef.current.length > 0) {
+      const matchedNames = new Set();
+      const avail = (pred) => pendingPreviewsRef.current.find(e => !matchedNames.has(e.name) && pred(e));
+      const updates = Object.fromEntries(newFiles.flatMap(f => {
+        const entry =
+          avail(e => e.name === f.filename) ??
+          avail(e => { const s = e.name.replace(/\.[^.]+$/, ''); return s && f.filename.includes(s); }) ??
+          avail(() => true);
```
santipalenque: on multiple uploads at same time, this can potentially match the wrong file. Worth a comment at least

src/components/progressive-img/index.js
priscila-moneo: Yes, I know. This change is intentional. The useRef is shared across all effect runs, so when src changes, the cleanup sets isCancelled.current = true on the shared ref — and the new effect's onload sees true and bails out immediately, leaving the image stuck in the blurred placeholder state forever. The let cancelled closure variable is scoped to each individual effect run, so each one starts fresh with false.

src/components/inputs/upload-input-v3/index.js
priscila-moneo: Fair point — while it's an edge case (same name + same size is essentially the same file), I've replaced dataURL from the thumbnail event with URL.createObjectURL(file) in addedfile. Each File object gets a unique blob URL by reference, so collisions are impossible regardless of name or size. As a bonus, the preview is now available immediately when the file is selected instead of waiting for Dropzone to generate the thumbnail. Blob URLs are revoked on cancel/error to avoid memory leaks.

src/components/inputs/upload-input-v3/index.js
priscila-moneo: useLayoutEffect runs synchronously before the browser paints, so when value updates, the completed uploading row is removed before the user sees anything. With useEffect it runs after paint, causing a one-frame flash where the file appears twice — once as "Complete" in the uploading section and once in the committed value section.

santipalenque: this just hides the symptom but does not address the re-render . which is caused by [this](https://github.com/OpenStackweb/openstack-uicore-foundation/pull/263#discussion_r3373823496)

priscila-moneo: Done the refactor you mention and prevValueRef, pendingPreviewsRef and the batched matching are all gone. But useLayoutEffect is still needed though: when value updates, both the completed uploading row and the new value row land in the same render, and without it the user sees them both for one frame before the cleanup runs. It's not hiding a symptom, it's preventing a real visual duplicate.

priscila-moneo: I'm working on the last of your comments and will update this PR with the latest code, will tag you so you can review it.

santipalenque (approved):
LGTM, thank you @priscila-moneo

smarcet:
Memory leak: blob URLs not revoked on unmount

`UploadInputV3` creates blob URLs in two places but has no cleanup when the component unmounts:

1. **`filePreviews` state** — blob URLs are stored here (keyed by server filename) in `wrappedOnUploadComplete`. They are only revoked in `handleRemove` (user-initiated delete). If the user navigates away from the page while uploaded files are showing, those blob URLs persist until page reload.

2. **`uploadingFiles[*].previewUrl`** — blob URLs created in `handleAddedFile`. If the component unmounts mid-upload, these are never revoked.

In a long-running SPA session where the user uploads images across multiple visits to this component without a full page reload, each unmount leaks the memory for all previewed images.

**Suggested fix** — add a cleanup effect using a ref to capture latest state at unmount time:

```js
const filePreviewsRef = useRef(filePreviews);
filePreviewsRef.current = filePreviews;
const uploadingFilesRef = useRef(uploadingFiles);
uploadingFilesRef.current = uploadingFiles;

useEffect(() => {
  return () => {
    Object.values(filePreviewsRef.current).forEach(url => URL.revokeObjectURL(url));
    uploadingFilesRef.current.forEach(f => f.previewUrl && URL.revokeObjectURL(f.previewUrl));
  };
}, []);
```

(Using refs instead of deps on `[]  ` because a cleanup-on-unmount effect with `[]` captures stale initial state, while refs always reflect the latest values.)

smarcet:
Zero/undefined `size` breaks preview mapping and renders misleading file size

Two issues compound each other when a file's `size` is `0` or `undefined`:

**1. `formatFileSize` uses a falsy guard (pre-existing, not introduced by this PR)**

```js
const formatFileSize = (bytes) => {
  if (!bytes) return '0 KB'; // catches 0, undefined, null, NaN
  ...
};
```

A file whose `size` is missing from the server response silently displays `0 KB` instead of something explicit like `—` or `Unknown`. Minor UX issue on its own.

**2. The size-based preview matching in `wrappedOnUploadComplete` has no defense against `size === 0` or `size === undefined` — introduced by this PR**

```js
const entry = prev.find(f => f.size === response.size && f.previewUrl);
```

If `response.size` comes back as `0` or `undefined` from the server, this `find` will match **any** uploading file that also has `size === 0` or `size === undefined` and has a preview URL. In a multi-file upload where the server omits the size field or where two empty files are uploaded concurrently, all of them collapse to the same match — the first entry wins and every file gets assigned the same (wrong) blob URL.

**Why these two are connected:** the `!bytes` guard in `formatFileSize` exists precisely because `size: 0` is a valid value that reaches the component. That same zero/undefined value is then used as an identity key for preview mapping with no validation.

**Suggested fix:** guard the preview mapping against falsy sizes before attempting the match:

```js
const entry = response?.size
  ? prev.find(f => f.size === response.size && f.previewUrl)
  : null;
```

This ensures that a missing or zero size from the server never triggers a spurious preview assignment.

src/components/inputs/upload-input-v3/index.js
smarcet: @priscila-moneo any  update on this one ?

smarcet (changes requested):
@priscila-moneo please re review

src/components/inputs/upload-input-v3/index.js
priscila-moneo: code already updated

smarcet (approved):
LGTM

PR #266, opened 2026-06-11
fix: add ellipsis tooltip component on mui tables

tomrndom:
ref: https://app.clickup.com/t/86bacrq3e

image

## Summary by CodeRabbit

* **New Features**
  * Added a reusable text truncation component for tables, supporting optional character limits and hover tooltips that display the full value.
  * Enabled truncation rendering within editable and sortable table cells.
* **Bug Fixes**
  * Improved consistent rendering for boolean cells and row action controls alongside custom cell content.
  * Refined table cell styling to better handle long text.
* **Tests**
  * Added/expanded coverage for truncation and ellipsis behavior, including tooltip display on overflow.


santipalenque (commented):
hey @tomrndom this looks good but I think we can make it a little bit better if we call the col prop truncateText and  allow passing true when you want to just use the col width, or a number of characters . wdyt ?

tomrndom:
> hey @tomrndom this looks good but I think we can make it a little bit better if we call the col prop truncateText and allow passing true when you want to just use the col width, or a number of characters . wdyt ?

@santipalenque I'll rename the col prop to truncateText. 

~~About the option to pass a number of characters I think this could be limited to only content that is a string to not mess with jsx code that is used in some cases.~~

~~If a number is passed but the content is jsx, it'll fall back to CSS-based truncation automatically.~~

I took a closer look at this, and it seems a bit messy to add a character number to truncate since the rendered content is always wrapped in a html span tag, this would required to pass to the ellipsis tooltip component new props as raw content and if the column has a render function, internally get the result of the render function, evaluate if it's a string and then truncate the text to display and get the tooltip with the full content.

I think that with the current approach using the width of the column we can achieve a similar result without making this component more complex and condition the posibble render based on the content of the cell

let me know what you think

santipalenque (approved):
LGTM

smarcet (approved):
LGTM

PR #278, opened 2026-07-01
Feature: new sponsor cart grid

santipalenque:
https://app.clickup.com/t/9014802374/86ban7rdx

## Summary by CodeRabbit

* **New Features**
  * Added expandable row details in the form table with an info/status indicator, inline notes, and a “Details” label.
* **Bug Fixes**
  * Row expansion and indicator state now follow Formik validation (touched/errors) instead of separate notes/settings callbacks.
* **Improvements**
  * Enhanced form inputs: required “*” markers, better label/slotProps handling, improved dropdown value coercion, and configurable checkbox margin.
* **Tests**
  * Updated and expanded test coverage for expansion/details and validation-driven behavior.
* **Chores**
  * Bumped the beta package version.


src/components/mui/FormItemTable/index.js
```diff
@@ -23,31 +25,93 @@ import {
...
+      });
+      return { ...prev, ...updates };
+    });
+  }, [errors, touched]);
+
+  const toggleRow = (rowId) => {
+    setOpenRows((prev) => ({ ...prev, [rowId]: !prev[rowId] }));
+  };
+
+  const getDetailsIconColor = (row) => {
```
smarcet, line 126: @santipalenque Form-class `Quantity` fields (e.g. "Qty of People") used to render as always-visible table columns; now they only live inside the collapsed `ExpandedRowContent` panel. The only cue telling a user to expand the row is this icon turning `"error"` red, which depends on `mf.is_required` being set on that specific metafield. In the test fixtures, the Form-class Quantity field doesn't have `is_required` set — if that also holds for real data, a user could leave a pricing-driving field empty, see only a neutral `"warning"` icon, and submit a row that silently totals $0.00.

Worth confirming `is_required` is reliably set on these fields server-side, or adding an explicit visual cue on the read-only Qty cell pointing to "Details".

src/components/mui/FormItemTable/index.js
```diff
@@ -23,31 +25,93 @@ import {
...
+            (f) => `i-${row.form_item_id}-c-${f.class_field}-f-${f.type_id}`
+          ),
+          `i-${row.form_item_id}-c-global-f-notes`
+        ]);
+        const hasVisibleError = Object.keys(errors).some(
+          (key) => expandedKeys.has(key) && touched[key]
+        );
+        if (hasVisibleError) updates[row.form_item_id] = true;
+      });
+      return { ...prev, ...updates };
```
smarcet: @santipalenque This spread always returns a new object reference even when `updates` is empty, so every time this effect fires it triggers a re-render of the whole table via `setOpenRows`. Since `errors`/`touched` get new references on most Formik validation cycles (default `validateOnChange`), this effect likely runs on most keystrokes while any row has an error, forcing a full-table re-render each time.

Consider returning `prev` unchanged when `Object.keys(updates).length === 0`.

src/components/mui/FormItemTable/__tests__/FormItemTable.test.js
```diff
@@ -561,66 +533,61 @@ describe("FormItemTable Component", () => {
...
             timeZone="America/New_York"
-            onNotesClick={mockOnNotesClick}
-            onSettingsClick={mockOnSettingsClick}
           />
         );
       }).not.toThrow();
     });
   });
 
   describe("ITEM Class Fields", () => {
```
smarcet, line 605: @santipalenque None of the updated tests cover the new `getDetailsIconColor` state machine (error/success/warning) or the auto-expand-on-validation-error effect in `FormItemTable/index.js` — only icon presence/toggle counts are asserted. Since these drive real form-completion signals, it'd be good to add a test that touches a required field with a validation error and asserts both the icon color and that the row auto-expands.

smarcet (changes requested):
@santipalenque please review

santipalenque:
done @smarcet

src/components/mui/FormItemTable/index.js
```diff
@@ -23,31 +24,93 @@ import {
...
+          ),
+          ...itemFields.map(
+            (f) => `i-${row.form_item_id}-c-${f.class_field}-f-${f.type_id}`
+          ),
+          `i-${row.form_item_id}-c-global-f-notes`
+        ]);
+        const hasVisibleError = Object.keys(errors).some(
+          (key) => expandedKeys.has(key) && touched[key]
+        );
+        if (hasVisibleError) updates[row.form_item_id] = true;
```
smarcet: @santipalenque This auto-expand effect can fight the user's manual collapse. `errors`/`touched` are whole-form Formik objects that get new object references on virtually every validation cycle (default `validateOnChange`), regardless of which field actually changed. Since `updates[row.form_item_id] = true` is set unconditionally whenever a row still has a visible error — without checking `prev[row.form_item_id]` first — a row the user just closed via `toggleRow` gets forced back open the next time the effect re-runs, which happens on the next keystroke in *any* other field/row, not just this one.

I reproduced this with a controlled harness (render with an error → auto-expand fires → manually collapse the row → pass a new `errors`/`touched` object with identical content, simulating Formik's next validation cycle) and the row snaps back open even though nothing about its own error changed.

Worth tracking rows the user explicitly dismissed (or only auto-expanding once per error occurrence) instead of recomputing `updates` from scratch on every effect run.

smarcet (changes requested):
@santipalenque please review
https://github.com/OpenStackweb/openstack-uicore-foundation/pull/278#discussion_r3508513200

src/components/mui/FormItemTable/index.js
santipalenque: @smarcet maybe this is a deeper discussion, what happens if a form item has 2 quantity extra fields (overriding global quantity) but one of them is not mandatory ? The calculated quantity will be 0

santipalenque: cc @JpMaxMan in case you want to chime in

JpMaxMan: It is hard for me to visualize this without seeing it - this is another case for being able to publish this to preview branches so design can review as well.  If I'm following correctly - yes we need a clear indicator that a required field is not being filled out.  Yes, it needs to be clear to the user what their optional fields are.  It's ok if it is zero if they chose not to fill it out.   Maybe the form expands as they are entering quantity or other visible data data then collapses when they move to the next item. Again, hard to visualize or even articulate without seeing it.

JpMaxMan: This also goes back to the idea of use what you build - have you looked in the old sponsor services on how they use this and how the interface worked @santipalenque ?

santipalenque: This is the issue @JpMaxMan : what happens if a form item has 2 quantity extra fields (overriding global quantity) but one of them is not mandatory ? The calculated quantity will be 0. 
You can visualize this in dev branch with the old designs, the issue is in the business rule, not the design.

JpMaxMan: In that case can you point me to a form in the dev environment that I can look at that demonstrates the issue.  I will look at it and I will make a video and I will share it with you and the design team.  But again, the point is for a sponsor to be able to clearly communicate what they need in terms of services and items for their booth.  The business rules and interface simply need to allow them to do that without tripping them up, looking broken, or collecting bad data.  Ideally I would love for you to be able to put yourself in the mind of the user and build something that accomplishes that.

santipalenque: yes I'm trying and most of the times I figure it out myself, but when I can't figure it out I ask for help ... I thought maybe you would know what is expected in this scenario, but don't worry I discussed it with Seba

JpMaxMan: I can't visualize the scenario - again if you could share with me a form on dev that depicts it then it would be helpful.  Maybe it is a hum dinger.  I'd just be surprised given that this interface is being brought over from an app that has been used by sponsors for many years.   Still if you can in any way show me, I'd appreciate it. Maybe you could make a JAM and link it here.  Or a link to a form in dev that is setup with this config.

smarcet:
@santipalenque `hasDrivingQuantityField` locks the row's global quantity field (and defaults the row open) whenever it finds an Item-class metafield of type `Quantity`:

```js
const hasItemLevel = (row.meta_fields ?? []).some(
  (mf) => mf.class_field === SPONSOR_FORMS_METAFIELD_CLASS.ITEM && mf.type === "Quantity"
);
return hasFormLevel || hasItemLevel;
```

But `calculateQuantity` in `index.js` only ever derives the computed value from `extraColumns` (Form-class fields) — it never reads Item-class fields, which I think is correct, since only Form-class quantities are meant to drive the row's quantity.

So for a row that has an Item-class `Quantity` field but no Form-class one, the global quantity input ends up forced `readOnly`/greyed out, nothing ever computes a value into it, and the user has no way to set it manually either — the row is stuck with whatever `initialValues` happened to provide.

Since only Form-class fields should drive the row quantity, I'd just drop the `hasItemLevel` branch from `hasDrivingQuantityField` so it matches what `calculateQuantity` actually does.

Could we add a regression test for this in `helpers.test.js` (there's no coverage for `hasDrivingQuantityField` at all right now)? Something like:

```js
describe("hasDrivingQuantityField", () => {
  test("returns false for a row whose only Quantity field is Item-class", () => {
    const row = {
      form_item_id: 1,
      meta_fields: [
        { type_id: 1, class_field: "Item", type: "Quantity" }
      ]
    };
    expect(hasDrivingQuantityField(row, [])).toBe(false);
  });

  test("returns true when a Form-class Quantity field exists", () => {
    const row = { form_item_id: 1, meta_fields: [] };
    const extraColumns = [{ type_id: 1, class_field: "Form", type: "Quantity" }];
    expect(hasDrivingQuantityField(row, extraColumns)).toBe(true);
  });
});
```

(note: `helpers.test.js` currently mocks `../../../../utils/constants` down to just `MILLISECONDS_IN_SECOND`, so the mock would need `SPONSOR_FORMS_METAFIELD_CLASS: { FORM: "Form", ITEM: "Item" }` added too for this to import cleanly.)

smarcet (changes requested):
@santipalenque please review https://github.com/OpenStackweb/openstack-uicore-foundation/pull/278#issuecomment-4916866923

smarcet (approved):
LGTM

PR #279, opened 2026-07-01
fix: new drag-n-drop-list component using dnd-kit lib

tomrndom:
ref: https://github.com/fntechgit/sponsor-services/pull/57

from https://github.com/fntechgit/sponsor-services/pull/57#discussion_r3505871793

## Summary by CodeRabbit

* **New Features**
  * Added a new drag-and-drop list component for MUI-based UIs, letting users reorder items visually.
  * Supports custom item rendering, drag handles, and automatic order updates after reordering.
  * Added a bundled entry so the component is available in builds.

* **Tests**
  * Added coverage for rendering, drag-and-drop reordering, missing IDs, custom ID fields, and order updates.

### DIFF
--- a/src/components/mui/DragNDropList/index.js
+++ b/src/components/mui/DragNDropList/index.js
@@ -24,39 +24,12 @@ import {
 import {
   SortableContext,
   sortableKeyboardCoordinates,
-  useSortable,
   verticalListSortingStrategy,
   arrayMove
 } from "@dnd-kit/sortable";
-import { CSS } from "@dnd-kit/utilities";
 import { Box } from "@mui/material";
 
-const SortableItem = ({ id, item, index, renderItem }) => {
-  const {
-    attributes,
-    listeners,
-    setNodeRef,
-    transform,
-    transition,
-    isDragging
-  } = useSortable({ id });
-
-  return (
-    <Box
-      ref={setNodeRef}
-      style={{
-        transform: CSS.Transform.toString(transform),
-        transition: [transition, "background 0.2s ease"].filter(Boolean).join(", ")
-      }}
-      sx={{ background: isDragging ? "#f0f0f0" : "inherit" }}
-    >
-      {renderItem(item, index, {
-        isDragging,
-        dragHandleProps: { ...listeners, ...attributes }
-      })}
-    </Box>
-  );
-};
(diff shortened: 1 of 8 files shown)


src/components/mui/DragNDropList/index.js
```diff
@@ -0,0 +1,132 @@
...
+    >
+      {renderItem(item, index, {
+        isDragging,
+        dragHandleProps: { ...listeners, ...attributes }
+      })}
+    </Box>
+  );
+};
+
+const getItemId = (item, index, idKey) =>
```
smarcet, line 42: @tomrndom Items without an `idKey` value fall back to a positional id (`new-${index}`), recomputed from the item's current index on every render. Since the same set of fallback ids persists across renders, after a drag the id that used to identify one item now identifies whatever item lands at that index — React reuses the component instance/fiber for that key instead of remounting it, so any local/uncontrolled state or transition inside `renderItem` can end up attached to the wrong logical item after a reorder.

This is a live path in the known consumer (sponsor-services `ExtraQuestionValuesFieldArray`), where newly pushed values have no `id` yet. It doesn't corrupt data there because the Formik fields are bound by `values[${index}].value` (positional, not identity-based), but it's worth documenting the caveat for other consumers whose `renderItem` carries local state — a stable client-generated id (e.g. `crypto.randomUUID()`) for new items would avoid the identity reuse entirely.

smarcet (changes requested):
@tomrndom please review

src/components/mui/drag-n-drop-list.js
```diff
@@ -0,0 +1,140 @@
...
+  SortableContext,
+  sortableKeyboardCoordinates,
+  useSortable,
+  verticalListSortingStrategy,
+  arrayMove
+} from "@dnd-kit/sortable";
+import { CSS } from "@dnd-kit/utilities";
+import { Box } from "@mui/material";
+
+const SortableItem = ({ id, item, index, renderItem }) => {
```
santipalenque: maybe this should go in a separate file inside components dir

santipalenque (approved):
other than this, LGTM

smarcet (approved):
LGTM

PR #282, opened 2026-07-03
feat: add order invoice component

tomrndom:
ref: https://app.clickup.com/t/9014802374/86baq9key

## Summary by CodeRabbit

* **New Features**
  * Added printable order invoices in PDF format with order details, customer/venue info, line items, fees, discounts, payments, refunds, notes, and calculated totals.
  * Supports both direct downloads and in-browser preview, including reconciliation and cancellation/refund impacts.
* **Tests**
  * Added a comprehensive Jest suite validating invoice row generation and the PDF download/preview flows.
* **Maintenance**
  * Upgraded the PDF rendering library and included the new order-invoice PDF bundle in the build output.


src/components/index.js
```diff
@@ -26,6 +26,7 @@ export {default as Input} from './inputs/text-input'
 export {default as Panel} from './sections/panel'
 export {default as SimpleLinkList} from './simple-link-list'
 export {default as SummitDropdown} from './summit-dropdown'
+export {OrderPdf, buildRows as buildOrderPdfRows, generateInvoicePDF, previewPDF} from './order-invoice-pdf'
```
smarcet: this is wrong please do check
https://github.com/OpenStackweb/openstack-uicore-foundation/blob/27283ec2d9676d7be4aad9c808b593a008c785f5/src/components/index.js#L134
this export force to incluide a 3rd party deps even when is not used

package.json
```diff
@@ -120,7 +120,7 @@
         "@mui/icons-material": "^6.4.3",
         "@mui/material": "^6.4.3",
         "@mui/x-date-pickers": "^7.26.0",
-        "@react-pdf/renderer": "^3.1.11",
+        "@react-pdf/renderer": "^3.1.11 || ^4.0.0",
```
smarcet: @tomrndom this is wrong u should use the same version on both places 
 "@react-pdf/renderer": "^4.4.1",

smarcet (changes requested):
@tomrndom please review

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,820 @@
...
+    typeBadge: {
+      icon: "DoNotDisturb",
+      iconColor: "#9ca3af",
+      label: "Cancelled",
+      labelColor: "#9ca3af"
+    }
+  }
+});
+
+const PdfTableRow = ({ row, styles, rowStyles }) => {
```
santipalenque: can you put each of the components and svgs in this file in its own file ? is very hard to read a 1000 line file

santipalenque (changes requested):

package.json
```diff
@@ -31,7 +31,7 @@
         "@mui/icons-material": "^6.4.3",
         "@mui/material": "^6.4.3",
         "@mui/x-date-pickers": "^7.26.0",
-        "@react-pdf/renderer": "^3.1.11",
+        "@react-pdf/renderer": "^4.4.1",
```
smarcet, line 34: @tomrndom `@react-pdf/renderer` is bumped from `^3.1.11` to `^4.4.1` here (and at line 129, peerDependencies), but `yarn.lock` wasn't updated — it still resolves `3.1.12`. This is a major, breaking upstream release (v4.0.0 dropped CJS support per the official changelog), and it's a peer dependency already used by `schedule-print-button.js`/`schedule-print-view.js`, which have no test coverage. Since this PR's own test suite fully mocks `@react-pdf/renderer`, the v3→v4 upgrade is currently unverified anywhere in the repo. Could we regenerate the lockfile and do a manual smoke test of `schedule-print` (and this new component) rendering under v4 before merging?

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,207 @@
...
+    throw error;
+  }
+};
+
+export const previewPDF = async (order, summit, { logoSrc, theme } = {}) => {
+  const blob = await pdf(
+    <OrderPdf order={order} summit={summit} logoSrc={logoSrc} theme={theme} />
+  ).toBlob();
+  const url = URL.createObjectURL(blob);
+  window.open(url, "_blank", "noopener,noreferrer");
```
smarcet: @tomrndom `window.open` here runs after `await pdf(...).toBlob()`, so the popup blocker no longer sees it as tied to the original click. Safari in particular will reliably block this (same pattern reported here: https://github.com/langgenius/dify/issues/29389). A common fix is to open a blank tab synchronously in the click handler (`window.open('', '_blank')`) and set `.location.href` once the blob URL is ready.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,207 @@
...
+    link.click();
+    URL.revokeObjectURL(url);
+    document.body.removeChild(link);
+  } catch (error) {
+    console.error("Error generating invoice PDF:", error);
+    throw error;
+  }
+};
+
+export const previewPDF = async (order, summit, { logoSrc, theme } = {}) => {
```
smarcet, line 255: @tomrndom Two small inconsistencies vs. `generateInvoicePDF` right above: this function has no try/catch, so a missing `order`/`summit` (the `OrderPdf: ... is required` throw) becomes an unhandled promise rejection here instead of a caught/logged error. Also, `generateInvoicePDF` revokes the object URL synchronously right after `link.click()`, while this one waits 60s — the immediate-revoke pattern has documented cases of producing empty/failed downloads on Safari and some Firefox versions (https://github.com/Scrince/Arculus_Recovery/issues/30). Worth aligning both on the same timing/error handling.

src/components/order-invoice-pdf/__tests__/order-invoice-pdf.test.js
```diff
@@ -0,0 +1,411 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import { buildRows, OrderPdf } from "../index";
```
smarcet: @tomrndom The suite covers `buildRows` and `<OrderPdf>`'s render output well, but never exercises `generateInvoicePDF`/`previewPDF` — the actual generate/preview feature this PR adds, and where the two issues above live. A test mocking `document.createElement`/`URL.createObjectURL`/`window.open` would catch both.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,195 @@
...
+  return cents < 0 ? `-${abs}` : abs;
+};
+
+export const buildRows = (order, summit) => {
+  const rows = [];
+  let balanceCents = 0;
+
+  (order.forms || []).forEach((form) => {
+    (form.items || [])
+      .filter((item) => item.quantity)
```
smarcet: @tomrndom This was flagged by CodeRabbit on Jul 3 ("Keep the quantity filter consistent with the default quantity") and marked resolved/addressed, but the underlying issue is still here: `.filter((item) => item.quantity)` drops items where `quantity` is `null`/`undefined`, yet line 108 defaults `qty` to `item.quantity ?? 1` — a default that can now never trigger, since anything reaching that point already passed the truthy filter. Items with a missing (not zero) quantity silently disappear from the invoice instead of rendering with an assumed qty of 1. A fresh Copilot review comment today flagged the same line independently, which confirms this regressed (likely during the file-split refactor) rather than being fixed. Suggest filtering with an explicit check like `(item.quantity ?? 1) > 0` so the default and the filter agree.

src/components/order-invoice-pdf/styles.js
```diff
@@ -0,0 +1,330 @@
...
+    marginBottom: 8
+  },
+  fieldRow: {
+    fontSize: 8,
+    lineHeight: 1.8,
+    flexDirection: "row",
+    color: "rgba(0, 0, 0, 0.87)",
+    paddingBottom: 3,
+    paddingTop: 6,
+    borderBottom: "1px solid rgba(224, 224, 224, 1)",
```
smarcet, line 75: @tomrndom Re: the Copilot comment on this line claiming `fieldRowNoBorder`'s `borderBottomWidth: 0` can't override this `borderBottom` shorthand — I traced `@react-pdf/stylesheet`'s actual resolution pipeline (`mergeStyles` → `expand`) in both v3.1.12 (installed locally) and v4.2.0 (same major line as this PR's dependency bump), and the code is identical in both. `mergeStyles` flattens the style array in order, and since `borderBottomWidth` doesn't exist in `fieldRow`, it gets appended as a new key after `borderBottom`. `expand()` then iterates keys in that same insertion order — so `borderBottom`'s shorthand expansion runs first, and the explicit `borderBottomWidth: 0` from `fieldRowNoBorder` is processed afterward, overwriting it. The override does work correctly here; no change needed on this line.

src/components/order-invoice-pdf/components.js
```diff
@@ -0,0 +1,142 @@
...
+      ) : (
+        <View style={styles.colType} />
+      )}
+      <View style={styles.colDescMultiLine}>
+        <Text style={s.descText}>
+          {row.type === "item"
+            ? `${row.description} - Total: ${row.qty}`
+            : row.description}
+        </Text>
+        {row.cancelledBy ? (
```
santipalenque: can't you just do rox.cancelledBy && <Text>.... ?  I think is cleaner than a ternary

src/components/order-invoice-pdf/components.js
```diff
@@ -0,0 +1,142 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * */
+
+import React from "react";
+import { View, Text, Svg, Path } from "@react-pdf/renderer";
+import { currencyAmountFromCents } from "../../utils/money";
+import { MUI_ICON_PATHS, PDF_ICON_SIZE, fmtBalance } from "./helpers";
+
+export const PdfIcon = ({ name, color, size = PDF_ICON_SIZE }) => (
```
santipalenque: each one of these

src/components/order-invoice-pdf/components.js
```diff
@@ -0,0 +1,142 @@
...
+    width={size}
+    height={size}
+    viewBox="0 0 24 24"
+    style={{ marginRight: 3 }}
+  >
+    <Path d={MUI_ICON_PATHS[name]} fill={color} />
+  </Svg>
+);
+
+export const FieldRow = ({ styles, label, value, noBorder = false }) => (
```
santipalenque: and this

src/components/order-invoice-pdf/components.js
```diff
@@ -0,0 +1,142 @@
...
+  <View style={[styles.fieldRow, noBorder && styles.fieldRowNoBorder]}>
+    <Text style={styles.fieldLabel}>{label ?? ""}</Text>
+    <Text style={styles.fieldValue}>{value ?? ""}</Text>
+  </View>
+);
+
+export const PdfTableRow = ({ row, styles, rowStyles }) => {
```
santipalenque: and this

src/components/order-invoice-pdf/components.js
```diff
@@ -0,0 +1,142 @@
...
+    </View>
+  );
+};
+
+// Uses order-level totals matching the table's ReconciliationBox props:
+//   cancelledTotal  ← order.cancelled_total
+//   refundsTotal    ← order.refunds_total
+//   retained        ← order.retained
+//   credited        ← order.credited_to_payment_method
+export const ReconciliationBlock = ({
```
santipalenque: etc etc.  should be its own file. 

Best Practices & Community ConsensusModern React development largely favors a pragmatic balance over rigid adherence to the rule:[1](https://www.reddit.com/r/reactjs/comments/12edlnp/how_strict_is_the_one_component_per_file_rule/). Define One "Public" Exported ComponentThe primary, default exported component should dominate the file. The file name should ideally match this component name (e.g., UserProfile.tsx).[2](https://www.reddit.com/r/reactjs/comments/y5wek2/should_i_create_a_separate_component_for_this/). Allow Small "Private" Sub-ComponentsIf a component needs a tiny helper component that won't be used anywhere else in the application, it is acceptable to write it in the same file. Keep it below the main component and do not export it.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,195 @@
...
+ * limitations under the License.
+ * */
+
+import moment from "moment-timezone";
+import { currencyAmountFromCents } from "../../utils/money";
+import { MILLISECONDS_IN_SECOND } from "../../utils/constants";
+
+export const DEFAULT_FONT_FAMILY = "Helvetica";
+
+export const MUI_ICON_PATHS = {
```
santipalenque: this would go in PdfIcon file

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,231 @@
...
+  if (!order) throw new Error("OrderPdf: order is required");
+  if (!summit) throw new Error("OrderPdf: summit is required");
+
+  const fontFamily = getThemeFontFamily(theme);
+  const styles = createStyles(fontFamily);
+  const rowStyles = createRowStyles(styles);
+  const rows = buildRows(order, summit);
+
+  const {
+    amount_due: total = 0,
```
smarcet: @tomrndom Quick question on `total` vs `amount_due`: `OrderPdf` reads `order.amount_due` here for the AMOUNT DUE line, but `SponsorOrderGrid` (`src/components/mui/SponsorOrderGrid/index.js:86`) reads `order.total` for the same concept. I checked `purchases-api`'s `purchase_v2_serializer.py` — the raw v2 response has `amount_due` (computed as `max(raw_amount - discount_amount + charges_amount - payments_total, 0)`) but no `total` field at all, so `amount_due` matches the backend contract exactly, same as the other 4 reconciliation fields this component already reads (`cancelled_total`, `refunds_total`, `retained`, `credited_to_payment_method`).

Can you confirm whether the `order` prop this component receives is the raw purchases-api v2 response, or whether it goes through the same mapping/normalization used to build `SponsorOrderGrid`'s `order` prop (which would carry `total` instead)? If it's the latter, AMOUNT DUE will always render $0.00 in production and `amount_due` should be `total` here.

smarcet (changes requested):
@tomrndom please review

src/components/order-invoice-pdf/index.js
tomrndom: This takes the raw `amount_due` prop directly from the order API response where total not exists, based on this change at sponsor-services https://github.com/fntechgit/sponsor-services/pull/100/changes

This components is migrated from sponsor-services, and reused part of the structure from the order-details page. On a deep look, that page the data comes from the reducer after a normalizeOrder function that assign the prop `total` from the value of `amount_due` or `net_amount` (https://github.com/fntechgit/sponsor-services/blob/main/src/utils/methods.js#L92), We should probably  need to adjust this line to support both options like the reducer

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,183 @@
...
+      });
+
+    const discountCents = form.discount_in_cents ?? 0;
+    if (discountCents) {
+      balanceCents -= discountCents;
+      rows.push({
+        rowKey: `discount-${form.id}`,
+        type: "discount",
+        code: "DIS",
+        description: String(form.discount || ""),
```
smarcet: @tomrndom `buildRows` reads `form.discount` as an already-formatted display string, but the real integration passes the raw, un-normalized order — so every discounted line renders with a blank "Details" description.

Confirmed against `fntechgit/summit-admin#1022`, which wires this component up: `handleInvoiceDownload` (`sponsor-purchases-tab/index.js` / `show-purchase-list-page/index.js`) calls `getSponsorOrder(...).then(({response: fetchedOrder}) => generateInvoicePDF(fetchedOrder, currentSummit, ...))` — `fetchedOrder` is the raw API response, never passed through `normalizeOrder()` (`summit-admin/src/pages/sponsors/sponsor-page/utils.js:6-13`), which is what actually produces `form.discount` (`formatDiscount(form.discount_amount, form.discount_type)`) for the on-screen grid. The raw payload only carries `discount_amount`/`discount_type` (see `summit-admin/.../edit-form/index.js:159-160`, which reads those two raw fields directly) — there's no raw `discount` field. So `String(form.discount || "")` resolves to `""` for every real order, and the invoice shows a "DIS" row with the correct amount but no indication of what the discount was.

This component's own tests mask the gap: the fixture passes `discount: "15%"` as a literal pre-formatted string, so nothing exercises the raw `discount_amount`/`discount_type` shape the actual caller sends.

Suggest deriving the description from `discount_amount`/`discount_type` directly (mirroring `formatDiscount` from `openstack-uicore-foundation/lib/utils/money`, which this same file already imports `currencyAmountFromCents` from) instead of assuming a pre-formatted `form.discount` string — that removes an implicit "caller must normalize first" contract this component doesn't document or enforce.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,237 @@
...
+  const {
+    net_amount,
+    amount_due,
+    cancelled_total: cancelledTotal = 0,
+    refunds_total: refundsTotal = 0,
+    retained = 0,
+    credited_to_payment_method: credited = 0
+  } = order;
+
+  const total = net_amount || amount_due || 0;
```
smarcet: @tomrndom `total` re-derives `net_amount || amount_due || 0` independently instead of sharing the one existing implementation of this rule.

Checked this against the actual integration in `fntechgit/summit-admin#1022`: `handleInvoiceDownload` passes the raw (non-normalized) order straight to `generateInvoicePDF`, so reading `net_amount`/`amount_due` directly here is actually necessary — not a bug by itself, since `normalizeOrder()` (`summit-admin/src/pages/sponsors/sponsor-page/utils.js:8`) uses the identical `data.net_amount || data.amount_due || 0` formula on the same raw fields. Both currently agree.

The risk is that this exact business rule now has two independent, hand-copied implementations across two repos (`summit-admin`'s `normalizeOrder` and this file), with nothing keeping them in sync. If either changes — field-priority fix, a new fallback field, a rounding/currency change — the on-screen "Amount Due" (`SponsorOrderGrid`/`TotalFooter`, which read the already-normalized `order.total`) and the downloaded invoice PDF can silently disagree on the same order.

Since `openstack-uicore-foundation` already owns the shared money helpers both repos import (`currencyAmountFromCents`, `formatDiscount`), this order-total derivation seems like a natural fit for the same shared location, so both the grid and the PDF compute "amount due" from one source instead of two.

smarcet (changes requested):
@tomrndom please review

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+  */
+export const getOrderTotal = (order) => {
+  if (!order) return 0;
+  return order.net_amount || order.amount_due || 0;
+};
+
+export const formatAddress = (address) => {
+  if (!address) return "";
+  return [
+    address.address_1,
```
smarcet: @tomrndom `formatAddress` reads `address_1`/`address_2`, but the client address coming from purchases-api v2 uses `line1`/`line2`, so the street is silently dropped from the invoice.

`PurchaseV2Serializer.get_address` projects `{line1, line2, city, state, postal_code, country}`, and `getSponsorOrder` (sponsor-services `src/actions/sponsor-actions.js:238`) returns `payload.response` from `/api/v2/.../purchases/{id}` unnormalized straight into `generateInvoicePDF` (`my-orders/index.js:66`). With a sponsor whose address is `123 Expo Blvd, Vancouver, BC, V6B 1A1, Canada`, the Client → Address block prints `Vancouver, BC, V6B 1A1, Canada`. The `.filter(Boolean)` added in 22ea1d8 removes the empty slots too, so the output looks like a valid address rather than a truncated one. `client_address_line1` is a populated field — `purchase_service.py:379` fills it from sponsor metadata and `:684` from the client-address edit endpoint.

The precedent is in the same app: `my-orders/order-details/index.js:55` renders `${address.line1} ${address.line2} ,\n${address.postal_code} ...`. The `zip_code ?? postal_code` fallback already on line 56 here handles exactly this shape mismatch for the postal code; `line1`/`line2` were missed.

Keep the summit-api keys first, since this same helper also formats the summit venue location (`index.js:110`), which genuinely uses `address_1`/`zip_code`:

```js
address.address_1 ?? address.line1,
address.address_2 ?? address.line2,
```

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+
+// Reads fontFamily off the consumer's MUI theme so it stays in sync with the
+// rest of their app instead of being duplicated as a separate literal. Only
+// the first name in the CSS font stack is usable here: react-pdf renders
+// with fonts registered via Font.register, so this name must match a family
+// already registered that way in the consuming app.
+export const getThemeFontFamily = (theme) => {
+  const fontFamily = theme?.typography?.fontFamily;
+  if (!fontFamily) return DEFAULT_FONT_FAMILY;
+  return fontFamily.split(",")[0].trim().replace(/^['"]|['"]$/g, "");
```
smarcet: @tomrndom This can return a font family react-pdf refuses to render, and the failure mode is a thrown error, not the fallback the comment at `index.js:34-38` describes.

`FontStore`'s constructor (`@react-pdf/font/lib/index.browser.js:344-400`) seeds only `Helvetica`, `Courier` and `Times-Roman`; `getFont` (`:471`) throws `Font family not registered: <x>` for anything else. `@react-pdf/layout/lib/index.js:1683-1695` does build a `[family, 'Helvetica']` fallback array, but it maps `fontStore.getFont` over it, so the throw on the first entry propagates before Helvetica is ever tried. There is no rescue — `pdf().toBlob()` rejects and no PDF is produced.

Concretely: sponsor-services' MUI theme is `fontFamily: ["Roboto", "sans-serif"].join(",")` (`src/styles/CustomTheme.js:27`) while its PDF registers `Inter` (`order-pdf/index.js:38,46`). The first caller that passes `theme` — the documented purpose of the new prop — gets a hard failure. MUI's own default theme has the same problem, since its default stack starts with `Roboto`. The current `!fontFamily` guard on line 76 only covers a theme with no `fontFamily` at all, not an unregistered one.

```js
export const getThemeFontFamily = (theme) => {
  const fontFamily = theme?.typography?.fontFamily;
  if (!fontFamily) return DEFAULT_FONT_FAMILY;
  const resolved = fontFamily.split(",")[0].trim().replace(/^['"]|['"]$/g, "");
  return Font.getRegisteredFontFamilies().includes(resolved)
    ? resolved
    : DEFAULT_FONT_FAMILY;
};
```

Worth updating the comment at `index.js:34-38` too — it currently promises a fallback that does not exist for this case.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,236 @@
...
+  // .location.href once the blob URL is ready.
+  const newTab = window.open("", "_blank");
+
+  try {
+    const blob = await pdf(
+      <OrderPdf order={order} summit={summit} logoSrc={logoSrc} theme={theme} />
+    ).toBlob();
+    const url = URL.createObjectURL(blob);
+
+    if (newTab) {
```
smarcet, line 269: @tomrndom When the popup blocker returns `null`, `previewPDF` resolves successfully having done nothing observable, so the caller has no way to tell the user why.

The blob is generated, `URL.createObjectURL` runs, the `if (newTab)` guard skips the navigation, and the function resolves `undefined` — identical to the success path. A user with popups blocked for the site clicks Preview and nothing happens, with no error to catch and no message to show. The URL is then revoked 60s later, so the work is thrown away too. The test `"does not throw when the popup blocker returns null"` currently pins this as intended behavior.

`generateInvoicePDF` right above always delivers the file, so preview is the only one of the two entry points that can fail invisibly. Falling back to the download path keeps the feature working instead of dead-ending:

```js
if (newTab) {
  newTab.location.href = url;
  setTimeout(() => URL.revokeObjectURL(url), REVOKE_DELAY_MS);
  return;
}
URL.revokeObjectURL(url);
return generateInvoicePDF(order, summit, { logoSrc, theme });
```

If you'd rather leave the decision to the consumer, returning a discriminated result (`{ opened: false }`) instead of `undefined` also works — the point is that the caller can distinguish the two outcomes.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+        price: currencyAmountFromCents(discountCents),
+        balanceCents
+      });
+    }
+  });
+
+  (order.fees || []).forEach((fee) => {
+    balanceCents += fee.amount;
+    rows.push({
+      rowKey: `fee-${fee.id}`,
```
smarcet: @tomrndom `fee.id` doesn't exist on the v2 fee payload — `PurchaseV2Serializer.get_fees` emits `line_id`, `position`, `title` and `amount` only, so every fee row gets `rowKey: "fee-undefined"`.

Item rows on line 104 already handle this correctly with `item.line_id ?? item.id`; fee rows were missed. `purchase_service.py:75-88` and `:289-300` create one Fee line per `cart.extra_costs` entry, so two or more fees on a single purchase is reachable (e.g. a purchase repriced across two payments each carrying an extra cost), and all of them collide on the same key.

On the current path the practical impact is limited — `pdf()` mounts once and React renders every child on initial mount regardless, so you get the duplicate-key warning rather than a missing row. It becomes a real defect if the `pdf()` instance is ever reused via `updateContainer`, or if this row list gains a re-render.

```js
rowKey: `fee-${fee.line_id ?? fee.id}`,
```

`SponsorOrderGrid/index.js` has the same `fee-row-${fee.id}` bug — probably worth a separate fix rather than folding it into this PR.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,197 @@
...
+  return fontFamily.split(",")[0].trim().replace(/^['"]|['"]$/g, "");
+};
+
+export const fmtBalance = (cents) => {
+  if (cents == null) return "";
+  const abs = currencyAmountFromCents(Math.abs(cents));
+  return cents < 0 ? `-${abs}` : abs;
+};
+
+export const buildRows = (order, summit) => {
```
smarcet: @tomrndom `buildRows` is a second, independent implementation of the order ledger `SponsorOrderGrid` already owns, and the two have to stay in sync by hand.

`SponsorOrderGrid/index.js:38-69` (`mapOrderData`) and its `calculateBalance` call sites at `:174-300` apply the same rules this function does — item `+`, discount `−`, fee `+`, payment `−`, refund `+`, payments and refunds interleaved by `created`. I checked the two against each other and found no divergence with a user-visible consequence today, so this isn't a bug report. The concern is that these are the numbers a sponsor sees on screen and the numbers on their invoice: a future adjustment to one is unlikely to reach the other, and the failure would be a silent mismatch between the two surfaces rather than an error.

This is the case `patterns/show-admin/summit-admin-reuse-before-build.md` § Duplicate Mapping Tables calls out ("Two copies of the same mapping drift silently when one gets updated and the other doesn't").

react-pdf obviously can't reuse the MUI rendering, but the derivation is pure: extracting the row-and-balance computation into one helper that both `SponsorOrderGrid` and `buildRows` consume would leave only presentation per renderer. Reasonable to defer to a follow-up if you'd rather not grow this PR — worth an explicit decision either way.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,236 @@
...
+  return (
+    <Document>
+      <Page size="A4" orientation="portrait" style={styles.page}>
+        {/* Row 1: Logo | Invoice title + Order / Date */}
+        <View style={styles.headerRow} wrap={false}>
+          <View style={styles.cellLeft}>
+            {logoSrc ? <Image src={logoSrc} style={styles.logo} /> : null}
+          </View>
+          <View style={styles.cellRight}>
+            <Text style={styles.invoiceTitle}>Invoice</Text>
```
smarcet: @tomrndom All user-visible copy in this component is hardcoded English, which makes the invoice the only part of the library a consumer can't localize or reword.

Around 25 strings across the component are literals: "Invoice", "Order", "Date", "Client", "Address", "Event", "Venue", "Pending", the table headers, "AMOUNT DUE", and in the subcomponents "Reconciliation", "Cancelled", "Refunded", "Retained as cancellation fee", "Credited to Payment Method", "Charge", "Payment", "Discount", "Refund", "NOTE", "Paid via", "Total:", and the `Cancelled by X on Y` sentence built in `helpers.js:95`.

`SponsorOrderGrid/components/ReconciliationBox.jsx:21-52` translates the identical four reconciliation labels via `T.translate("sponsor_order_grid.reconciliation" | ".cancelled" | ".refunded" | ".retained" | ".credited")`, and those keys already exist in `src/i18n/en.json`. `patterns/show-admin/summit-admin-i18n-pattern.md` § Key Placement is explicit: "Do not add inline English strings for user-facing UI unless the adjacent file already intentionally uses a literal technical value."

Suggest routing the copy through `T.translate` under an `order_invoice_pdf.*` namespace and adding the keys to `src/i18n/en.json`, reusing `sponsor_order_grid.*` for the labels the grid already defines.

smarcet (changes requested):
@tomrndom please review

src/components/order-invoice-pdf/__tests__/order-invoice-pdf.test.js
```diff
@@ -0,0 +1,629 @@
...
+
+// ─── OrderPdf render-level ────────────────────────────────────────────────────
+
+const makeRenderOrder = (overrides = {}) => ({
+  number: "ORD-2026-001",
+  total: 0,
+  purchased_date: 1700000000,
+  purchased_by_full_name: "Jane Doe",
+  client: { contact_name: "Jane Doe", company_name: "Acme Corp" },
+  address: null,
```
smarcet: @tomrndom The fixtures in this suite are hand-authored rather than captured, so the tests go green through exactly the payload-shape mismatches that keep coming back in review.

Two live examples in this file: `address: null` here means `formatAddress` is never exercised, so the `address_1` vs `line1` mismatch (separate thread on `helpers.js:52`) passes CI untouched. And the fee objects built in the `buildRows` tests carry an `id`, which `PurchaseV2Serializer.get_fees` never emits — it projects `line_id` — so the `fee-undefined` row key (separate thread on `helpers.js:137`) also passes CI. In both cases the fixture asserts a payload the API does not send.

This is the loop that's costing the PR its time, not those two bugs specifically. Twelve of the fifteen commits on this branch are `fix:` commits adjusting a single payload field discovered during review — "adjust order total from net_amount or amount_due values", "adjust main_location object based on summit data", "adjust form item quantity filter condition", "add trim on format address function". The review is doing the job the fixtures should be doing, and that has no natural stopping point: each round only fixes the field somebody happened to notice.

Suggest pinning the contract once, with a real response. Same call sponsor-services actually makes (`src/actions/sponsor-actions.js:231-234`):

```
GET {PURCHASES_API_URL}/api/v2/summits/{summit_id}/sponsors/{sponsor_id}/purchases/{id}
    ?expand=forms,forms.items,forms.items.meta_fields,forms.items.type,refunds,payments,notes,fees
```

Pick an order that exercises the shape — populated client address, two or more fees, a cancelled item, a refund, a discounted form — scrub names and emails, commit it as `__tests__/fixtures/purchase-v2.json`, and build each test case as an override on top of it instead of a hand-written object. That converts an open-ended review loop into a finite checklist: any field the component reads that the API doesn't send becomes a failing assertion instead of a silently empty cell in the PDF.

Worth sharing the same fixture with `SponsorOrderGrid`'s tests, which hand-author the reconciliation totals the same way (`SponsorOrderGrid.test.js:160-162`) and read the same endpoint's payload.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,236 @@
...
+            />
+          </View>
+        </View>
+
+        {/* Row 2: Client + Address | Event + Venue + Address */}
+        <View style={styles.headerRow} wrap={false}>
+          <View style={styles.cellLeft}>
+            <FieldRow
+              styles={styles}
+              label="Client"
```
smarcet: @tomrndom Four fields that `order-details` shows for the same order don't appear anywhere on the invoice — flagging to confirm the omissions are deliberate rather than lost in the port.

`my-orders/order-details/index.js` renders these from the same raw v2 `currentOrder` this component receives:

- **Sponsor and tier** — `sponsor.company_name` (`:74-76`) and `sponsor.sponsorships.map((s) => s.type_name).join(", ")` (`:77-80`). The invoice never names the sponsor. The Client block here is the purchasing contact and their company, which isn't necessarily the same entity.
- **Payment method** — `currentOrder.payment_method` (`:90`).
- **Payment status** — `currentOrder.status` (`:94`).
- **Payment date** — `formatDate(currentOrder.created, "LOC", DATETIME_FORMAT)` (`:98`). This component has a single Date field, bound to `purchased_date`.

None of these is a defect on its own. The reason it's worth an explicit decision: an invoice that doesn't identify the sponsor, doesn't say how it was paid, and doesn't say whether it's paid is unusual for a document the sponsor files and may forward to their finance team. Right now the AMOUNT DUE line is the only signal of payment state, and on a fully settled order it renders `$0.00` in green rather than saying anything — easy to misread as "no charge" instead of "paid".

If the omissions are intentional, a line in the PR description would save the next reader from re-deriving this. If they aren't: `payment_method`, `status` and `created` are already on the payload and are field reads, but `sponsor` is not — `order-details` receives it as a separate prop and the purchase only carries `sponsor_id`, so that one would mean a new prop on `OrderPdf`.

smarcet: Follow-up on the four fields above, after tracing where each surface's order actually comes from.

`order-details` and this component receive different shapes of the same purchase. `sponsor-orders-list-reducer.js:69-71` runs `normalizeOrder(payload.response)` on `RECEIVE_SPONSOR_ORDER`, so the page renders the normalized order; `my-orders/index.js:66` passes the value the `getSponsorOrder` thunk returns, which is `payload.response` straight from the API. That difference is behind some of the field lookups this port had to rework, but it does **not** apply to these four — I re-checked each against the raw payload, which is the one this component receives.

It does split the list in two, though, and the answers are probably different:

- `payment_method`, `status` and `created` are all in `PurchaseV2Serializer.allowed_fields`, so they're already on the object handed to `generateInvoicePDF`. No new prop, no extra request — three field reads. For these, "lost in the port" now looks more likely to me than a deliberate scope decision, which is why I'd rather have your read than assume either way.
- `sponsor` and tier are a different question. The purchase only carries `sponsor_id`; `order-details` gets the sponsor object from `showAccessState.showAccess.sponsor` as a separate prop. Adding it here means widening `OrderPdf`'s signature, so it's an API decision rather than a field read — reasonable to defer out of this PR.

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,284 @@
...
+            <FieldRow
+              styles={styles}
+              label={T.translate("order_invoice_pdf.order")}
+              value={order.number}
+            />
+            <FieldRow
+              styles={styles}
+              label={T.translate("order_invoice_pdf.date")}
+              value={order.purchased_date ? formatDate(
+                order.purchased_date,
```
smarcet, line 87: @tomrndom The Date field renders `purchased_date` in `summit.time_zone_id`, but the screen this component ports renders the same field in the viewer's local time — `order-details/index.js:61-62` does `formatDate(currentOrder.purchased_date, "LOC", DATE_FORMAT)`, and Payment Date right below this one already follows that convention (`created` + `"LOC"`).

As-is the invoice mixes two timezones in adjacent header fields, and the Date shown on the PDF can disagree with the Date the same user just saw on the order-details page for the same order.

Suggested fix: use `"LOC"` here, matching `order-details` and the Payment Date field. (The `buildRows` payment/cancellation dates are net-new — the grid never renders them — so they should follow the same choice for a single-timezone document.)

src/components/order-invoice-pdf/index.js
```diff
@@ -0,0 +1,284 @@
...
+            {rows.map((row) => (
+              <PdfTableRow
+                key={row.rowKey}
+                row={row}
+                styles={styles}
+                rowStyles={rowStyles}
+              />
+            ))}
+
+            <ReconciliationBlock
```
smarcet: @tomrndom `ReconciliationBlock` renders unconditionally, but the grid this ports gates it behind the consumer-controlled `withReconciliation` prop (`SponsorOrderGrid/index.js:334`).

A clean order with nothing cancelled or refunded prints "Reconciliation — Cancelled $0.00 / Refunded $0.00 / Balance $0.00" directly above an AMOUNT DUE line that can read $600.00 — contradictory on a customer-facing financial document, and `OrderPdf` exposes no prop to suppress it.

Suggested fix: render the block only when any of `cancelledTotal`/`refundsTotal`/`retained`/`credited` is nonzero (or accept a `withReconciliation`-style prop with that default), restoring parity with `SponsorOrderGrid`.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+  * Derives an order's amount due directly from its raw net_amount/amount_due
+  * fields (both in cents). This is the single source for that fallback rule,
+  * shared by consumers that normalize a raw order (e.g. into order.total) and
+  * by consumers that render straight from the raw order (e.g. the invoice PDF).
+  * @param {object} order - Raw order object with net_amount/amount_due in cents.
+  * @returns {number} - The amount due, in cents.
+  */
+export const getOrderTotal = (order) => {
```
smarcet: @tomrndom `getOrderTotal`'s doc comment promises it's "the single source" for the `net_amount || amount_due` rule, shared by consumers that normalize raw orders — but it lives in a module whose top-level imports include `Font` from `@react-pdf/renderer`.

Any consumer importing just this money rule (e.g. a sponsor-services reducer computing `order.total`) drags the whole PDF stack into that bundle, so in practice the cross-repo dedup this enables won't happen. The established home for money helpers is `utils/money` — this file already imports `currencyAmountFromCents`/`formatDiscount` from there.

Suggested fix: move `getOrderTotal` (and arguably `fmtBalance`) to `src/utils/money.js` and re-import here.

src/components/order-invoice-pdf/helpers.js
```diff
@@ -0,0 +1,205 @@
...
+  ].sort((a, b) => a.created - b.created);
+
+  paymentsAndRefundsOrdered.forEach((item) => {
+    if (item._rowType === "payment") {
+      balanceCents -= item.amount;
+      rows.push({
+        rowKey: `payment-${item.id}`,
+        type: "payment",
+        code: T.translate("mui_table.pay"),
+        description: `${T.translate("mui_table.paid_via")} ${item.method || "card"}`,
```
smarcet: @tomrndom Last hardcoded user-visible string left after the i18n pass: the `"card"` fallback renders "Paid via card" with untranslatable English regardless of locale.

Suggested fix: move it behind a key, e.g.

```js
description: `${T.translate("mui_table.paid_via")} ${item.method || T.translate("mui_table.card")}`
```

with `"card": "card"` added to `mui_table` in `en.json`.

src/components/order-invoice-pdf/index.js
smarcet: @tomrndom Confirmed resolved in code as of 2f8c284: `getOrderTotal` (now at `src/utils/money.js:223`) implements `net_amount || amount_due || 0` — the exact rule `normalizeOrder` applies in sponsor-services (`src/utils/methods.js:92`), so the invoice and the order-details page now share the same amount-due rule. Remaining follow-up is on the sponsor-services side: `normalizeOrder` can import `getOrderTotal` from `openstack-uicore-foundation/lib/utils/money` to drop its own copy. Marking as resolved.

src/components/order-invoice-pdf/helpers.js
smarcet: @tomrndom Decision recorded: this stays as-is in this PR and the extraction goes to a follow-up ticket: https://app.clickup.com/t/86bb67u52. Scope there: a pure `buildOrderLedger(order)` in `src/utils/order-ledger.js` consumed by both `SponsorOrderGrid` and `buildRows`, leaving only presentation per renderer — it also fixes the three divergences that already exist between the two copies (quantity filter, `fee.id` row keys on v2 payloads, zero-amount discount rows) and drops `buildRows`' now-unused `summit` param. Resolving in favor of the follow-up.

smarcet (approved):
LGTM

PR #283, opened 2026-07-03
fix: add snackbarErrorMsg function to actions

tomrndom:
ref: https://app.clickup.com/t/9014802374/86baqq2rh

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -60,6 +60,12 @@ export const snackbarSuccessHandler = (message) => (dispatch, state) =>
         state
     );
 
+export const snackbarErrorMsg = (message) => (dispatch, state) =>
+    setSnackbarMessage({ ...message, type: "error", code: CODE_200 })(
+        dispatch,
+        state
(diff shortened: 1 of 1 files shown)


src/utils/actions.js
```diff
@@ -60,6 +60,12 @@ export const snackbarSuccessHandler = (message) => (dispatch, state) =>
         state
     );
 
+export const snackbarErrorMsg = (message) => (dispatch, state) =>
```
smarcet, line 63: @tomrndom this need to be exported by webpack config

smarcet (changes requested):
@tomrndom please review

src/utils/actions.js
tomrndom: @smarcet this is exported here https://github.com/OpenStackweb/openstack-uicore-foundation/blob/main/webpack.common.js#L163 with all the actions of the fiel

smarcet (approved):
LGTM

PR #289, opened 2026-07-13
fix: company-input-v2 — explicit selection only, no clear icon, "Use "<typed>"" row

JpMaxMan:
## The bug
`CompanyInputV2` used MUI's `autoSelect`, which commits the currently **highlighted** option on blur. Users would mouse over a suggestion, then tab away intending to keep their typed text — and the moused-over company got saved instead (wrong company populated).

## Changes (`src/components/inputs/company-input-v2.js`)
- **Remove `autoSelect`** — selection is now **explicit** (click / Enter) only. This is the root-cause fix.
- **New `onBlur`** — tab/click-away commits **exactly what was typed**: resolved to an existing company only on an *exact* (case-insensitive) name match, otherwise a free-text `{ id: 0, name }`. Never commits a merely-highlighted option.
- **`disableClearable`** — removes the MUI clear (x) icon (the field commits free text, so an explicit clear affordance isn't wanted; delete the text to empty).
- **`Use "<typed>"` row** — `filterOptions` appends a synthetic option so users can explicitly commit their typed text; skipped when it already matches a listed company. Commits a clean `{ id: 0, name }` (display-only marker stripped). `renderOption` shows the `Use "…"` label; `getOptionLabel` still returns the plain name.
- Predictive typeahead unchanged (`freeSolo` + server-side `queryRegistrationCompanies`).

## ⚠️ Shared-component behavior change
This affects **all** `CompanyInputV2` consumers, not just the requesting screen: no more clear (x) icon, and blur no longer auto-selects a highlighted option (it keeps the typed text). Both are intentional — flagging for reviewers.

Note: the Formik variant `CompanyInputMUI` (`mui/formik-inputs/company-input-mui.js`) is a separate implementation and is **not** touched here; if a Formik form needs the same behavior we'll do it in a follow-up.

## Tests
+5 tests; the file's suite is **25/25 green**:
- blur keeps the typed text and never commits a highlighted option (pins the bug);
- blur resolves an exact case-insensitive match to the canonical company;
- clear icon never renders, even with a real selected value;
- `Use "…"` row commits clean free text; no redundant `Use` row on an exact match.

Verified by running jest directly (`node_modules/.bin/jest`) — CI's Node 18 avoids a local `node-sass`/Node-25 build issue unrelated to this change. `package.json` version intentionally not bumped (done separately at release).

## Summary by CodeRabbit

* **New Features**
  * Added a synthetic “Use ‘…’” option to commit a company name that isn’t in the suggestions.
  * “Use ‘…’” is shown only when it doesn’t already match an available option (trimmed, case-insensitive).
* **Bug Fixes**
  * Leaving the field (including browser autofill) now correctly commits typed/free-text values when there’s no exact match; exact-name matches commit the existing option instead.
  * Removed the clear (x) icon to keep the selection behavior consistent and prevent redundant updates.
* **Tests**
  * Expanded blur/autofill/clear test coverage and updated assertions for the “Use ‘…’” row behavior.

### DIFF
--- a/src/components/inputs/__tests__/company-input-v2.test.js
+++ b/src/components/inputs/__tests__/company-input-v2.test.js
@@ -223,6 +223,34 @@ describe("CompanyInputV2 integration", () => {
         expect(committed).toEqual({ id: 0, name: "Autofilled Co" });
     });
 
+    it("clears the committed company when the field is emptied (delete-all-text) and blurred", () => {
+        // With disableClearable there's no (x), so deleting the text and blurring
+        // is the only way to clear — the value must propagate as null.
+        queryRegistrationCompanies.mockImplementation(() => {});
+
+        const onChange = jest.fn();
+        renderControlled({ initialValue: { id: 1, name: "Tipit" }, onChange });
+        const input = screen.getByRole("combobox");
+
+        fireEvent.change(input, { target: { value: "" } });
+        fireEvent.blur(input);
+
+        const committed = onChange.mock.calls.map((c) => c[0].target.value).pop();
+        expect(committed).toBeNull();
+    });
+
+    it("does not fire a redundant change when an already-empty field is blurred", () => {
+        queryRegistrationCompanies.mockImplementation(() => {});
+
+        const onChange = jest.fn();
+        renderControlled({ initialValue: null, onChange });
+        const input = screen.getByRole("combobox");
+
+        fireEvent.blur(input);
+
+        expect(onChange).not.toHaveBeenCalled();
+    });
+
     it("on blur commits the canonical existing match when the typed text matches case-insensitively", () => {
         // Capture the API callback so we can resolve it manually.
         let resolveQuery;
--- a/src/components/inputs/__tests__/company-input-v2.test.js
+++ b/src/components/inputs/__tests__/company-input-v2.test.js
@@ -203,6 +203,26 @@ describe("CompanyInputV2 integration", () => {
         expect(committed).toEqual({ id: 0, name: "Tip" });
     });
 
+    it("commits a browser-autofilled value on blur even when it never fired onInputChange (#241 iOS Chrome)", () => {
+        // Autofill writes straight to the DOM without React's onInputChange, so
+        // the component's input state stays empty. The blur handler must read
+        // the DOM value (event.target.value), not React state, or required-field
+        // validation regresses — the exact bug #241's autoSelect fixed.
+        queryRegistrationCompanies.mockImplementation(() => {});
+
+        const onChange = jest.fn();
+        renderControlled({ onChange });
+        const input = screen.getByRole("combobox");
+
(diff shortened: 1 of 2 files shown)


JpMaxMan:
### Re: `autoSelect` and the #241 autofill fix — addressed in `1bc112a`

@gcutrini flagged (via Slack) that `autoSelect` was added in #241 to fix a real bug: in `freeSolo` mode MUI only propagates a value on explicit selection, so **typing-and-tabbing and browser autofill (iOS Chrome)** never reached the parent and required-field validation failed. Correct — and simply removing `autoSelect` would regress that.

**Reconciliation (latest commit):** `autoSelect` stays removed — it's what committed the currently *highlighted* option on blur, so mousing over a suggestion and tabbing away saved the wrong company. In its place, `onBlur` now reads the input's **DOM value** (`event.target.value`) — the *same source* `autoSelect` read internally — and commits it as the field value. So:

- ✅ typed-and-tabbed values still propagate on blur (#241)
- ✅ **autofilled** values still propagate on blur, even when they never fire `onInputChange` (#241 — new test: *"commits a browser-autofilled value on blur even when it never fired onInputChange"*)
- ✅ a merely *highlighted/hovered* option is **never** committed (the bug this PR set out to fix)

Suite: 26/26 (jest run directly; CI Node 18 avoids a local node-sass/Node-25 issue).

⚠️ **Merge gate:** the autofill test passes in jsdom, but jsdom ≠ iOS Chrome. Reading `event.target.value` is strictly the more-robust source (it's what `autoSelect` used), but the real iOS-Chrome autofill path needs a device re-test. @gcutrini — could you verify that path on this branch, since you have the #241 repro? Happy to pair if the behavior differs.

JpMaxMan:
### CodeRabbit finding (clear-on-empty in `onBlur`) — addressed in `0f79e52`

Verified against the current code: **valid**, and worth fixing precisely because this PR removes the clear (x) icon (`disableClearable`), making delete-all-text + blur the only way to clear the field. Previously `onBlur` fired nothing for empty input, so the prior company stayed committed (a required field would still read as filled).

Fix (minimal):
- `onBlur` now propagates `null` when the field is emptied, guarded by `if (normalizedValue)` so an already-cleared field doesn't emit a redundant change.
- Non-empty exact-match / free-text behavior unchanged.
- Regression tests added: *"clears the committed company when the field is emptied (delete-all-text) and blurred"* and *"does not fire a redundant change when an already-empty field is blurred"*. Suite 28/28.

gcutrini:
Small UX nit: the `Use "<typed>"` row lands at the bottom. Feels better at the top. Typed text is the user's primary intent, and arrow-down-then-Enter would commit it without scrolling past matches. One-line change in `filterOptions`:

```js
return trimmed && !alreadyListed
  ? [{ id: 0, name: trimmed, isFreeTextOption: true }, ...opts]
  : opts;
```

smarcet (approved):
LGTM

PR #290, opened 2026-07-14
fix(company-input-v2): explicit selection, no clear icon, Use "<typed>" row (v4 backport)

gcutrini:

Backport of #289 to the v4.x line.

## The bug
`CompanyInputV2` used MUI's `autoSelect`, which commits the currently **highlighted** option on blur. Users would mouse over a suggestion, then tab away intending to keep their typed text — and the moused-over company got saved instead (wrong company populated).

## Changes (`src/components/inputs/company-input-v2.js`)
- **Remove `autoSelect`** — selection is now **explicit** (click / Enter) only. This is the root-cause fix.
- **New `onBlur`** — tab/click-away commits **exactly what was typed**: resolved to an existing company only on an *exact* (case-insensitive) name match, otherwise a free-text `{ id: 0, name }`. Never commits a merely-highlighted option. Reads the DOM value (`event.target.value`) so browser autofill (notably Chrome iOS) is captured even when React input state is stale — preserves the fix from #241.
- **`disableClearable`** — removes the MUI clear (x) icon (the field commits free text, so an explicit clear affordance isn't wanted; delete the text to empty).
- **`Use "<typed>"` row** — `filterOptions` **prepends** a synthetic option so users can explicitly commit their typed text; skipped when it already matches a listed company. Commits a clean `{ id: 0, name }` (display-only marker stripped).

## ⚠️ Shared-component behavior change
This affects **all** `CompanyInputV2` consumers, not just the requesting screen: no more clear (x) icon, and blur no longer auto-selects a highlighted option (it keeps the typed text). Both are intentional — flagging for reviewers.

## Tests
28/28 green on the file's suite. Coverage includes: hover-then-tab guard, Chrome iOS autofill DOM-value path, empty-blur clear, `disableClearable`, `Use "…"` row commit + hidden on exact match, canonical-case resolution on blur.

### DIFF
--- a/src/components/inputs/__tests__/company-input-v2.test.js
+++ b/src/components/inputs/__tests__/company-input-v2.test.js
@@ -178,6 +178,79 @@ describe("CompanyInputV2 integration", () => {
         return { ...utils, getValue: () => setValue };
     };
 
+    it("on blur takes the typed text and, when it matches no option, commits it as free text (never a highlighted option)", () => {
+        // The bug this guards: with autoSelect, mousing over a suggestion and
+        // tabbing away committed that highlighted company. Now blur must commit
+        // exactly what was typed.
+        let resolveQuery;
+        queryRegistrationCompanies.mockImplementation((_summitId, _input, cb) => {
+            resolveQuery = cb;
+        });
+
+        const onChange = jest.fn();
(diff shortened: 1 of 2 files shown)


smarcet (approved):
LGTM

PR #291, opened 2026-07-15
fix(company-input-v2): keep Use row on refocus, purge stale free-text (v4.x)

gcutrini:

Follow-up to #290 addressing two dropdown regressions found during local verification.

## Bugs fixed

**1. `Use "<typed>"` row disappeared after committing free-text and refocusing** — the `alreadyListed` check treated the just-committed `{id: 0, name}` entry as a real match and suppressed the Use row. Also affected passive refocus (click back with no typing) because MUI resets `params.inputValue` to empty when the display value matches the selection.

**2. Stale free-text option flashed briefly on refocus after a new commit** — the previously-committed `{id: 0}` option lingered in the options list until the debounced API response replaced it. Refocus in that window showed the stale entry.

## Changes

**Fixes (2 commits):**
- `filterOptions` now only counts real (id > 0) companies as "already listed" for the Use row check, and falls back to the committed free-text's name when `params.inputValue` is empty.
- Effect purges any leftover `{id: 0}` options synchronously the moment `normalizedValue` changes, before firing the API request.

**Refactors on top (8 commits, no behavior change):**
- Rename `findExistingByName` → `findExistingCompany` (mirrors `isExistingCompany`).
- Extract `namesMatch` — dedupes the `X.trim().toLowerCase()` comparison across three sites.
- Extract `shouldOfferUseRow` — unit-testable predicate for the Use row.
- Extract `resolveTypedCompany` — dedupes onBlur/onChange typed-string resolution.
- Simplify onBlur `currentName` via `getOptionName`.
- Extract `getUseRowText` — formalises the active-typing + committed-free-text fallback.
- Extract `resolveCommittedCompany`, use `fireChange` in onChange (removes prop/handler name shadow, dedupes with the rest of the component).
- Extract `findCanonicalUpgrade`, simplify effect's options builder.

## Naming convention

The extracted helpers follow a consistent verb-prefix convention already in the file: `is*` predicates, `find*` returns match or null, `resolve*` transforms input to a Company (always returns), `get*` computed getter, `should*` boolean decision.

## Tests

**34/34 green.** Each fix commit ships with an integration test that fails without the fix (verified by reverting each in isolation). The `shouldOfferUseRow` extraction adds four unit tests including the id:0-doesnt-suppress-Use case that pins the regression.

### DIFF
--- a/src/components/inputs/__tests__/company-input-v2.test.js
+++ b/src/components/inputs/__tests__/company-input-v2.test.js
@@ -17,7 +17,8 @@ import CompanyInputV2, {
     isExistingCompany,
     isNewCompany,
     findExistingCompany,
-    normalizeCompanyValue
+    normalizeCompanyValue,
+    shouldOfferUseRow
 } from "../company-input-v2";
 
 // Mock the API helper so tests can drive the callback synchronously.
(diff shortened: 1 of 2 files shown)


smarcet (approved):
LGTM

PR #297, opened 2026-07-23
Feature: lazy load reload

santipalenque:
https://app.clickup.com/t/9014802374/86bb1gp6z

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery for stale or unavailable application chunks by reloading the page.
  * Added safeguards to prevent repeated reload loops and improved reporting when recovery fails.
  * Added support for handling chunk-loading errors across lazy-loaded features and global application errors.

* **Tests**
  * Added comprehensive coverage for chunk detection, reload behavior, error handling, and recovery scenarios.

* **Chores**
  * Updated the package to version `5.0.45-beta.0`.

### DIFF
--- a/src/utils/__tests__/lazy-with-reload.test.js
+++ b/src/utils/__tests__/lazy-with-reload.test.js
@@ -1,0 +1,353 @@
+jest.mock("react", () => ({
+    ...jest.requireActual("react"),
+    lazy: jest.fn((loader) => loader),
+}));
+
+const mockCaptureMessage = jest.fn();
+jest.mock("@sentry/react", () => ({
+    captureMessage: (...args) => mockCaptureMessage(...args),
+}));
+
+import {
+    isChunkLoadError,
+    reloadOnChunkError,
+    lazyWithReload,
+    initChunkErrorRecovery,
+    chunkErrorSentryBeforeSend,
+} from "../lazy-with-reload";
+
+const chunkLoadError = () => Object.assign(new Error("Loading chunk 3 failed.\n(missing: https://app.test/foo_abc123.js)"), {
+    name: "ChunkLoadError",
+});
+
+const htmlParsedAsScriptError = (filename) => Object.assign(
+    new Error("Unexpected token '<'"),
+    { name: "SyntaxError", filename }
+);
+
+// Replaces document's <script> tags so getBuildFingerprint() (cached per
+// module instance) reads a specific, known set the next time a fresh
+// module instance computes it.
+const setDocumentScripts = (srcs) => {
+    document.querySelectorAll("script[data-test-build]").forEach((el) => el.remove());
+    srcs.forEach((src) => {
+        const script = document.createElement("script");
+        script.src = src;
+        script.setAttribute("data-test-build", "1");
+        document.body.appendChild(script);
+    });
+};
+
+beforeEach(() => {
+    window.sessionStorage.clear();
+    window.SENTRY_DSN = "https://test.example/dsn";
+    mockCaptureMessage.mockClear();
+    console.error = jest.fn(); // eslint-disable-line no-console
+    console.log = jest.fn(); // eslint-disable-line no-console
+    delete window.location;
+    window.location = { reload: jest.fn() };
+    setDocumentScripts([]);
+});
+
+describe("isChunkLoadError", () => {
+    test("matches webpack's ChunkLoadError by name, regardless of message", () => {
+        expect(isChunkLoadError(chunkLoadError())).toBe(true);
+    });
+
+    test("matches a SyntaxError for 'Unexpected token <' with no filename", () => {
+        expect(isChunkLoadError(htmlParsedAsScriptError())).toBe(true);
+    });
+
+    test("matches a SyntaxError for 'Unexpected token <' when the filename looks like a content-hashed chunk", () => {
+        expect(
+            isChunkLoadError(
+                htmlParsedAsScriptError(),
+                "https://app.test/static/dashboard_a1b2c3d4e5f6.js"
+            )
+        ).toBe(true);
+    });
+
+    test("does not match a SyntaxError for 'Unexpected token <' when the filename does not look like a chunk", () => {
+        expect(
+            isChunkLoadError(
+                htmlParsedAsScriptError(),
+                "https://app.test/static/vendor.min.js"
+            )
+        ).toBe(false);
+    });
+
+    test("does not match an unrelated error", () => {
+        expect(isChunkLoadError(new TypeError("x is not a function"))).toBe(false);
+    });
+
+    test("does not match null/undefined", () => {
+        expect(isChunkLoadError(null)).toBe(false);
+        expect(isChunkLoadError(undefined)).toBe(false);
+    });
+
+    test("accepts a custom chunk filename pattern", () => {
+        const customPattern = /^\/legacy-chunks\//;
+        expect(
+            isChunkLoadError(
+                htmlParsedAsScriptError(),
+                "/legacy-chunks/dashboard.js",
+                customPattern
+            )
+        ).toBe(true);
+        expect(
+            isChunkLoadError(
+                htmlParsedAsScriptError(),
+                "/static/dashboard_a1b2c3d4.js",
+                customPattern
+            )
+        ).toBe(false);
+    });
+});
+
+describe("reloadOnChunkError", () => {
+    test("reloads once and returns true for a first-time chunk error", () => {
+        const handled = reloadOnChunkError(chunkLoadError());
+
+        expect(handled).toBe(true);
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        expect(mockCaptureMessage).not.toHaveBeenCalled();
+    });
+
+    test("does not reload again on a second chunk error on the same build, and reports it instead", () => {
+        reloadOnChunkError(chunkLoadError());
+        window.location.reload.mockClear();
+
+        const handled = reloadOnChunkError(chunkLoadError());
+
+        expect(handled).toBe(false);
+        expect(window.location.reload).not.toHaveBeenCalled();
+        expect(mockCaptureMessage).toHaveBeenCalledWith(
+            "Chunk load error persisted after auto-reload",
+            expect.objectContaining({
+                tags: { chunkErrorRecovery: "failed" },
+            })
+        );
+    });
+
+    test("falls back to console.error when Sentry is not initialized", () => {
+        delete window.SENTRY_DSN;
+        reloadOnChunkError(chunkLoadError());
+        window.location.reload.mockClear();
+
+        reloadOnChunkError(chunkLoadError());
+
+        expect(mockCaptureMessage).not.toHaveBeenCalled();
+        expect(console.error).toHaveBeenCalledWith(
+            "Chunk load error persisted after auto-reload",
+            expect.objectContaining({
+                tags: { chunkErrorRecovery: "failed" },
+            })
+        );
+    });
+
+    test("does nothing for a non-chunk error", () => {
+        const handled = reloadOnChunkError(new TypeError("boom"));
+
+        expect(handled).toBe(false);
+        expect(window.location.reload).not.toHaveBeenCalled();
+        expect(mockCaptureMessage).not.toHaveBeenCalled();
+    });
+
+    // Reproduces the "two deploys in the same long-lived tab" scenario: a
+    // reload already happened once for build A, and now build B (a distinct
+    // set of statically-loaded <script> tags) has its own, unrelated chunk
+    // failure - it should get its own fresh reload attempt rather than
+    // being silently blocked by the earlier, unrelated recovery.
+    test("allows a fresh reload when a chunk error occurs on a different build than the one already reloaded for", () => {
+        jest.resetModules();
+        setDocumentScripts(["https://app.test/main_buildA111111.js"]);
+        // eslint-disable-next-line global-require
+        const buildA = require("../lazy-with-reload");
+        expect(buildA.reloadOnChunkError(chunkLoadError())).toBe(true);
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        window.location.reload.mockClear();
+
+        // still build A: same fingerprint, must not reload again
+        expect(buildA.reloadOnChunkError(chunkLoadError())).toBe(false);
+        expect(window.location.reload).not.toHaveBeenCalled();
+        mockCaptureMessage.mockClear();
+
+        // simulate the reload having carried the tab onto a new build
+        jest.resetModules();
+        setDocumentScripts(["https://app.test/main_buildB222222.js"]);
+        // eslint-disable-next-line global-require
+        const buildB = require("../lazy-with-reload");
+
+        expect(buildB.reloadOnChunkError(chunkLoadError())).toBe(true);
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        expect(mockCaptureMessage).not.toHaveBeenCalled();
+    });
+});
+
+describe("lazyWithReload", () => {
+    test("on a chunk-load failure, reloads and returns a promise that never resolves", async () => {
+        const importer = () => Promise.reject(chunkLoadError());
+        const loader = lazyWithReload(importer);
+
+        let settled = false;
+        loader().then(() => { settled = true; }, () => { settled = true; });
+        await new Promise((resolve) => setTimeout(resolve, 0));
+
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        expect(settled).toBe(false);
+    });
+
+    test("on an unrelated failure, rethrows so the error boundary still sees it", async () => {
+        const originalError = new TypeError("boom");
+        const importer = () => Promise.reject(originalError);
+        const loader = lazyWithReload(importer);
+
+        await expect(loader()).rejects.toBe(originalError);
+        expect(window.location.reload).not.toHaveBeenCalled();
+    });
+
+    test("on success, resolves normally", async () => {
+        const module = { default: () => null };
+        const importer = () => Promise.resolve(module);
+        const loader = lazyWithReload(importer);
+
+        await expect(loader()).resolves.toBe(module);
+    });
+});
+
+describe("initChunkErrorRecovery", () => {
+    // tracked so afterEach can always remove the listeners, even if a test's
+    // own assertions throw first - otherwise a failing test would leak
+    // listeners onto window and contaminate later tests in this file
+    let cleanup;
+
+    afterEach(() => {
+        cleanup?.();
+        cleanup = undefined;
+    });
+
+    test("reloads and prevents default on an unhandledrejection carrying a chunk error", () => {
+        cleanup = initChunkErrorRecovery();
+        const preventDefault = jest.fn();
+
+        window.dispatchEvent(
+            Object.assign(new Event("unhandledrejection"), {
+                reason: chunkLoadError(),
+                preventDefault,
+            })
+        );
+
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        expect(preventDefault).toHaveBeenCalledTimes(1);
+    });
+
+    test("reloads and prevents default on a global error event for the HTML-as-script SyntaxError", () => {
+        cleanup = initChunkErrorRecovery();
+        const preventDefault = jest.fn();
+
+        window.dispatchEvent(
+            Object.assign(new Event("error"), {
+                error: htmlParsedAsScriptError(),
+                filename: "https://app.test/static/dashboard_a1b2c3d4e5f6.js",
+                preventDefault,
+            })
+        );
+
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+        expect(preventDefault).toHaveBeenCalledTimes(1);
+    });
+
+    test("does not touch unrelated errors", () => {
+        cleanup = initChunkErrorRecovery();
+        const preventDefault = jest.fn();
+
+        window.dispatchEvent(
+            Object.assign(new Event("error"), {
+                error: new TypeError("boom"),
+                preventDefault,
+            })
+        );
+
+        expect(window.location.reload).not.toHaveBeenCalled();
+        expect(preventDefault).not.toHaveBeenCalled();
+    });
+
+    test("cleanup removes the listeners", () => {
+        cleanup = initChunkErrorRecovery();
+        cleanup();
+        cleanup = undefined;
+        const preventDefault = jest.fn();
+
+        window.dispatchEvent(
+            Object.assign(new Event("unhandledrejection"), {
+                reason: chunkLoadError(),
+                preventDefault,
+            })
+        );
+
+        expect(window.location.reload).not.toHaveBeenCalled();
+    });
+
+    test("logs a confirmation once when a prior reload was pending, and clears the marker", () => {
+        window.sessionStorage.setItem("uicore.chunk-load-error-pending-confirmation", "1");
+
+        cleanup = initChunkErrorRecovery();
+
+        expect(console.log).toHaveBeenCalledWith(
+            "[lazy-with-reload] Recovered from a chunk load error via automatic reload"
+        );
+        expect(
+            window.sessionStorage.getItem("uicore.chunk-load-error-pending-confirmation")
+        ).toBeNull();
+    });
+
+    test("does not log a confirmation when no reload was pending", () => {
+        cleanup = initChunkErrorRecovery();
+
+        expect(console.log).not.toHaveBeenCalled();
+    });
+
+    test("a real reload-then-init cycle logs exactly one confirmation", () => {
+        // first "page load": a chunk error triggers the reload, which sets
+        // the pending marker (sessionStorage survives a real reload)
+        reloadOnChunkError(chunkLoadError());
+        expect(window.location.reload).toHaveBeenCalledTimes(1);
+
+        // "post-reload" bootstrap calls initChunkErrorRecovery once, as the
+        // app does in app.js
+        cleanup = initChunkErrorRecovery();
+
+        expect(console.log).toHaveBeenCalledWith(
+            "[lazy-with-reload] Recovered from a chunk load error via automatic reload"
+        );
+        expect(console.log).toHaveBeenCalledTimes(1);
+    });
+});
+
+describe("chunkErrorSentryBeforeSend", () => {
+    test("drops an event whose exception is a ChunkLoadError", () => {
+        const event = {
+            exception: { values: [{ type: "ChunkLoadError", value: "Loading chunk 3 failed" }] },
+        };
+        expect(chunkErrorSentryBeforeSend(event)).toBeNull();
+    });
+
+    test("drops an event whose exception is the HTML-as-script SyntaxError", () => {
+        const event = {
+            exception: { values: [{ type: "SyntaxError", value: "Unexpected token '<'" }] },
+        };
+        expect(chunkErrorSentryBeforeSend(event)).toBeNull();
+    });
+
+    test("passes through an unrelated exception unchanged", () => {
+        const event = {
+            exception: { values: [{ type: "TypeError", value: "x is not a function" }] },
+        };
+        expect(chunkErrorSentryBeforeSend(event)).toBe(event);
+    });
+
+    test("passes through a message-type event with no exception unchanged", () => {
+        const event = { message: "Chunk load error persisted after auto-reload" };
+        expect(chunkErrorSentryBeforeSend(event)).toBe(event);
+    });
+});
--- a/src/utils/lazy-with-reload.js
+++ b/src/utils/lazy-with-reload.js
@@ -1,0 +1,224 @@
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+  const message = (error && error.message) || (typeof error === "string" ? error : "");
+  // "SyntaxError" can show up on error.name (a real SyntaxError object) or
+  // inline in the message text (window.onerror's flat event.message string,
+  // e.g. "Uncaught SyntaxError: Unexpected token '<'") - check both.
+  const looksLikeSyntaxError =
+    (error && error.name === "SyntaxError") || /SyntaxError/i.test(message);
+  if (!looksLikeSyntaxError || !/Unexpected token '<'/i.test(message)) {
+    return false;
+  }
+  return !filename || chunkFilenamePattern.test(filename);
```
smarcet: @santipalenque The filename-less branch of the SyntaxError heuristic turns unrelated JSON.parse failures into page reloads and hides them from Sentry.

`return !filename || chunkFilenamePattern.test(filename)` matches any SyntaxError whose message contains "Unexpected token '<'" when no filename is available — and the two call sites that pass no filename are exactly the ones where unrelated errors arrive. Verified concrete scenario (V8/Chromium): `new Response('<html>502</html>').json()` rejects with `SyntaxError: Unexpected token '<', "<html>502</html>" is not valid JSON`. If that rejection goes unhandled (a fetch of an API that returned an HTML 502/proxy/auth page), `onUnhandledRejection` (line 189) passes it here with `filename === undefined`, and the user's page does a full reload mid-session — once per build, with the next occurrence filing a false "Chunk load error persisted after auto-reload" report. The same filename-less call in `chunkErrorSentryBeforeSend` (line 218) permanently drops this whole class of errors from Sentry monitoring.

Tightening this loses no genuine webpack case: webpack's JSONP runtime sets `error.name = "ChunkLoadError"` on every chunk-load failure it detects through the import() promise, including the "script responded 200 with HTML but the chunk never registered" case (webpack/lib/web/JsonpChunkLoadingRuntimeModule.js, loadingEnded handler, errorType 'missing') — so the ChunkLoadError path already covers the stale-chunk failure this branch is trying to catch.

Suggested fix:
- In `onUnhandledRejection`, match only `ChunkLoadError` (import() rejections are never raw SyntaxErrors).
- In `isHtmlParsedAsScriptError`, require a filename that matches `chunkFilenamePattern` (drop the `!filename ||` escape) — the genuine HTML-as-script error arrives via the global `error` event with `event.filename` set.
- In `chunkErrorSentryBeforeSend`, extract the culprit filename from the exception's stack frames and pass it through (as the existing CodeRabbit thread on the beforeSend hook suggests).
- Update the test "matches a SyntaxError for 'Unexpected token <' with no filename" (lazy-with-reload.test.js line 57), which currently pins the permissive behavior, and add a regression test asserting a JSON.parse-style SyntaxError does NOT trigger a reload via unhandledrejection.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+      typeof document !== "undefined"
+        ? Array.from(document.scripts)
+            .map((s) => s.src)
+            .filter(Boolean)
+            .sort()
+            .join(",")
+        : "";
+  }
+  return cachedBuildFingerprint;
+};
```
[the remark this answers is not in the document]
smarcet: @santipalenque Confirmed after verification — this one is real and worth fixing. Two additions to the bot's analysis:

The header comment (lines 61-68) claims the fingerprint captures the scripts "present the first time this module runs", but the cache is populated on first *call*, and the only call site is inside `reloadOnChunkError` (line 127) — at first-failure time, after webpack's runtime has appended script tags for any lazy chunks that loaded successfully. So the stored fingerprint includes a variable chunk set.

Concrete impact when the reload doesn't fix things (stale cached index.html — exactly the scenario the throttle exists for): fingerprint at failure #1 = static scripts + chunks loaded pre-deploy; after reload, failure #2's fingerprint = static scripts only → mismatch → a second reload the throttle should have blocked, and the "persisted after auto-reload" report is delayed one full cycle. The loop is bounded (it stops as soon as a fingerprint repeats), so the typical damage is one extra reload plus delayed failure telemetry — but it defeats the module's core "once per build" guarantee.

The suggested eager-at-module-evaluation fix is correct, with the caveat CodeRabbit already noted: it relies on this module being evaluated at bootstrap, before any lazy import — which holds for consumers that call `initChunkErrorRecovery` at app start.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+
+  writeReloadState({ build: currentBuild });
+  try {
+    window.sessionStorage.setItem(PENDING_CONFIRMATION_KEY, "1");
+  } catch {
+    // ignore - see readReloadState/writeReloadState
+  }
+  window.location.reload();
+  return true;
+};
```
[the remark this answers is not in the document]
smarcet: @santipalenque Confirmed after verification — real, and likely the *common* case rather than an edge case. A route rendering several `lazyWithReload` components mounts them together, so a stale build produces two or more import() rejections in the same microtask drain. `window.location.reload()` doesn't halt JS: the first rejection writes RELOAD_STATE_KEY synchronously (line 135) and triggers the reload; the second rejection's handler still runs, reads the state just written (line 130), sees the same build, and files "Chunk load error persisted after auto-reload" — a false recovery-failure alert on a recovery that is actually in flight. It also returns false, so `lazyWithReload` rethrows into the error boundary for a visible flash before navigation.

Since this fires on most multi-chunk recoveries, the recovery-failed Sentry signal — the one thing this module emits for monitoring — becomes mostly noise. The suggested in-memory `reloadTriggeredThisPage` guard is the right shape; please also add a test covering two same-tick failures (currently untested).

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+  // "SyntaxError" can show up on error.name (a real SyntaxError object) or
+  // inline in the message text (window.onerror's flat event.message string,
+  // e.g. "Uncaught SyntaxError: Unexpected token '<'") - check both.
+  const looksLikeSyntaxError =
+    (error && error.name === "SyntaxError") || /SyntaxError/i.test(message);
+  if (!looksLikeSyntaxError || !/Unexpected token '<'/i.test(message)) {
+    return false;
+  }
+  return !filename || chunkFilenamePattern.test(filename);
+};
```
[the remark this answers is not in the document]
smarcet: Not applicable as stated. webpack's JSONP runtime sets error.name = "ChunkLoadError" on every chunk-load failure it detects through the import() promise — including the "script responded 200 with HTML but the chunk never registered" case (webpack/lib/web/JsonpChunkLoadingRuntimeModule.js, loadingEnded handler, errorType 'missing') — and it does so in every browser. So in Firefox the stale-chunk failure still triggers the safety-net reload via the ChunkLoadError path; the SyntaxError message heuristic is a redundant second net for the same failure, not the only trigger. Broadening the message match would only widen a heuristic that already over-matches (see the JSON.parse false-positive discussion on the beforeSend thread). Resolving this thread.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+export const chunkErrorSentryBeforeSend = (event) => {
+  const exceptionValues = event.exception?.values || [];
+  const matchesChunkError = exceptionValues.some((exceptionValue) =>
+    isChunkLoadError({
+      name: exceptionValue.type,
+      message: exceptionValue.value
+    })
+  );
+  return matchesChunkError ? null : event;
+};
```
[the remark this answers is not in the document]
smarcet: @santipalenque Confirmed after verification — real, and narrower than the full problem. The filename-less `isChunkLoadError` call here (line 218) means any SyntaxError whose message contains "Unexpected token '<'" is silently dropped from Sentry, not just stale-chunk errors. Verified concrete case (V8/Chromium): `Response.json()` on an HTML error page (proxy 502, auth redirect) rejects with `SyntaxError: Unexpected token '<', "<html>..." is not valid JSON` — a class of production failures this filter would hide from monitoring permanently.

CodeRabbit's suggested fix (extract the culprit filename from the exception's stack frames and pass it through) is the right shape. Note the same filename-less gap also exists in `onUnhandledRejection` (line 189), where it causes page reloads rather than just suppressed events — see the full analysis and combined fix on the line 54 thread.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+  return !filename || chunkFilenamePattern.test(filename);
+};
+
+export const isChunkLoadError = (error, filename, chunkFilenamePattern) =>
+  isWebpackChunkLoadError(error) ||
+  isHtmlParsedAsScriptError(error, filename, chunkFilenamePattern);
+
+// Fingerprints the statically-loaded <script> tags present the first time
+// this module runs on a given page load - i.e. before any lazy chunk has
+// been requested. This identifies "which build is currently running"
```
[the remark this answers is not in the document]
santipalenque: false

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,224 @@
...
+) => {
+  const message = (error && error.message) || (typeof error === "string" ? error : "");
+  // "SyntaxError" can show up on error.name (a real SyntaxError object) or
+  // inline in the message text (window.onerror's flat event.message string,
+  // e.g. "Uncaught SyntaxError: Unexpected token '<'") - check both.
+  const looksLikeSyntaxError =
+    (error && error.name === "SyntaxError") || /SyntaxError/i.test(message);
+  if (!looksLikeSyntaxError || !/Unexpected token '<'/i.test(message)) {
+    return false;
+  }
```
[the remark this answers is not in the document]
santipalenque: fixed

src/utils/__tests__/lazy-with-reload.test.js
```diff
@@ -0,0 +1,353 @@
...
+beforeEach(() => {
+    window.sessionStorage.clear();
+    window.SENTRY_DSN = "https://test.example/dsn";
+    mockCaptureMessage.mockClear();
+    console.error = jest.fn(); // eslint-disable-line no-console
+    console.log = jest.fn(); // eslint-disable-line no-console
+    delete window.location;
+    window.location = { reload: jest.fn() };
+    setDocumentScripts([]);
+});
```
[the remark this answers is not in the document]
santipalenque: the window.location claim is empirically false in this repo's actual
  test runtime, and the console-mock point, while technically true, is a minor style nit on pre-existing code (not anything we touched), not something causing test failures or incorrect
  behavior. Nothing here is blocking or risky enough to warrant a change.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,230 @@
...
+// a message-type event (no .exception), so it is unaffected by this filter.
+// Sentry stack frames are ordered oldest-first; the last one is where the
+// exception was actually thrown - the only frame worth checking against
+// chunkFilenamePattern.
+const getExceptionFilename = (exceptionValue) => {
+  const frames = exceptionValue.stacktrace?.frames;
+  return frames?.length ? frames[frames.length - 1]?.filename : undefined;
+};
+
+export const chunkErrorSentryBeforeSend = (event) => {
```
smarcet, line 231: @santipalenque In production builds, a chunk failure from a bare `React.lazy()` inside an error boundary never reaches `initChunkErrorRecovery`, and once this `beforeSend` is wired it also disappears from Sentry — a silent, unrecovered failure.

Mechanism: React 17's `lazy` attaches its own rejection handler to the `import()` promise (`react/cjs/react.development.js:1280-1300`), so the failure is never an `unhandledrejection`; it is rethrown during render and absorbed by the nearest boundary. summit-admin wraps the whole app in `Sentry.ErrorBoundary` (`src/app.js:183`) and keeps ~10 bare `React.lazy` routes (`src/layouts/primary-layout.js:22-26`, `sponsor-id-layout.js:23-36`). For those, the user gets the fallback with no reload, the boundary's `captureReactException` (`@sentry/react` `errorboundary.js:59-71`) produces an exception event with `type: "ChunkLoadError"`, and this hook drops it. Note this only shows in production: in dev, React's `invokeGuardedCallbackDev` re-dispatches the error as a real `window` `error` event (`react-dom.development.js:3989`), so local testing would see the reload fire and mask the gap. The comment at lines 176-178 claiming bare `React.lazy()` is covered is not accurate for this topology.

Suggested fix:
- Only drop a chunk event here when a reload is actually in flight, e.g. export a `isChunkReloadPending()` reading `reloadTriggeredThisPage` and `return matchesChunkError && isChunkReloadPending() ? null : event;` — unrecovered chunk failures then still reach Sentry.
- Correct the 176-178 comment and add a note (module header or consumer docs) that existing `React.lazy` calls must migrate to `lazyWithReload` to get recovery; `initChunkErrorRecovery` alone does not cover them behind a boundary.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,230 @@
...
+  const currentBuild = getBuildFingerprint();
+  const state = readReloadState();
+
+  if (state && state.build === currentBuild) {
+    reportRecoveryFailed(error, filename);
+    return false;
+  }
+
+  reloadTriggeredThisPage = true;
+  writeReloadState({ build: currentBuild });
```
smarcet: @santipalenque The once-per-build throttle fails open when `sessionStorage` is unavailable, which turns a persistently failing chunk into an unbounded reload loop with no `reportRecoveryFailed` ever filed.

`readReloadState` returns `null` and `writeReloadState` swallows the throw (lines 72-87), while `reloadTriggeredThisPage` resets on every page load — so each load sees "no prior reload for this build" and reloads again. Reproduced against this module with a throwing `sessionStorage` stub: 5 simulated page loads → 5 `window.location.reload()` calls, 0 reports. The precondition is narrow (browsers blocking all storage, sandboxed iframes without `allow-same-origin`, where the app's own localStorage-backed auth is already broken), but a throttle should fail closed, not open.

Suggested fix: have `writeReloadState` return `false` when it cannot persist, and in `reloadOnChunkError` skip the reload (call `reportRecoveryFailed` instead, or just return `false`) when the state could not be written:

```js
if (!writeReloadState({ build: currentBuild })) {
  reportRecoveryFailed(error, filename);
  return false;
}
```

Plus a test covering the storage-unavailable path.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,230 @@
...
+// boundary/fallback UI.
+export const lazyWithReload = (importer, chunkFilenamePattern) =>
+  React.lazy(() =>
+    importer().catch((error) => {
+      if (reloadOnChunkError(error, undefined, chunkFilenamePattern)) {
+        // reload is in flight - never resolve so nothing renders (e.g. an
+        // error boundary flash) before it takes effect
+        return new Promise(() => {});
+      }
+      throw error;
```
smarcet, line 182: @santipalenque When two `lazyWithReload` chunks fail in the same tick, the second one rethrows into the error boundary while the reload from the first is still in flight.

`reloadOnChunkError` returns `false` on the second call because `reloadTriggeredThisPage` is already set (line 122), so this branch throws; `React.lazy` marks the component rejected and the boundary fallback renders until the new document arrives (`window.location.reload()` does not halt JS or unload the current page immediately). Sentry noise is not an issue — the boundary's `ChunkLoadError` event is dropped by `chunkErrorSentryBeforeSend` — so the effect is a cosmetic fallback flash. Reach is limited today: summit-admin's layouts render lazies as mutually exclusive `<Route>`s inside a `<Switch>` (`src/layouts/primary-layout.js:66-106`), so it needs a page that mounts two or more `lazyWithReload` components at once.

Suggested fix: return the never-resolving promise whenever a reload is pending this page, not only when this call triggered it — e.g. `if (reloadOnChunkError(...) || isChunkReloadPending()) return new Promise(() => {});` (same helper as suggested on the `beforeSend` thread), with a test for two same-tick `lazyWithReload` failures asserting neither rejects.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+
+  const currentBuild = getBuildFingerprint();
+  const state = readReloadState();
+
+  if (state && state.build === currentBuild) {
+    reportRecoveryFailed(error, filename);
+    return false;
+  }
+
+  reloadTriggeredThisPage = true;
```
smarcet: @santipalenque `reloadTriggeredThisPage` is set before the write succeeds, so a failed `writeReloadState` leaves the page permanently in the "reload pending" state while no reload ever happens.

`isChunkReloadPending()` then returns true for the rest of the page lifetime. In `lazyWithReload` (:174-181) that makes the catch return `new Promise(() => {})`, so the component suspends forever: the user gets a permanent Suspense fallback with no reload and no error boundary, and `chunkErrorSentryBeforeSend` (:239) drops the chunk exceptions on top of it. Reproduced against this module with a `sessionStorage` stub whose `setItem` throws: 0 reload calls, the promise never settles, `isChunkReloadPending()` is true, and beforeSend returns null.

This contradicts the flag's own documented contract - the comment at :178-179 reads "reload is in flight", and the header at :105-108 says the function returns true "only when it actually triggers a *new* reload". On this path nothing is in flight. It also defeats the intent of the fail-closed change: the agreed behavior was to skip the reload and let the error propagate normally, which is exactly what does not happen through `lazyWithReload`.

On reachability, so this gets prioritized correctly: it needs `sessionStorage.setItem` to throw while `localStorage` still works, and I could not find a path where that holds today. `putOnLocalStorage` (src/utils/methods.js:133-137) has no try/catch and runs during the PKCE flow (security/methods.js:104 and :507), so in any storage-blocked environment auth throws before an authenticated user reaches a lazy route; sessionStorage quota exhaustion is implausible (summit-admin has zero sessionStorage usage). Still worth fixing - it is two lines, and it becomes reachable the moment `putOnLocalStorage` is hardened.

Suggested fix - set the flag only on a path that actually reloads:

```js
  if (!writeReloadState({ build: currentBuild })) {
    reportRecoveryFailed(error, filename);
    return false;
  }
  reloadTriggeredThisPage = true;
```

Plus the storage-unavailable test requested on the fail-closed thread, which has not landed: a throwing `setItem` asserting no reload and that the `lazyWithReload` promise rejects.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+    // ignore
+  }
+};
+
+// Drop-in replacement for React.lazy(() => import(...)) that auto-recovers
+// from a stale-chunk failure by reloading the page once. On any other error,
+// or a repeat chunk failure after the reload already happened once for this
+// build, the error is rethrown so it still reaches the app's own error
+// boundary/fallback UI.
+export const lazyWithReload = (importer, chunkFilenamePattern) =>
```
smarcet: @santipalenque The `chunkFilenamePattern` parameter is inert here - `lazyWithReload` can never act on it.

Line 175 always passes `undefined` as the filename, and the only branch that consults the pattern (`isHtmlParsedAsScriptError`, :36-48) short-circuits on `Boolean(filename)` since the filename cross-check was added. The pattern is therefore matched against nothing. Verified: passing `/.*/` (matches any string) together with a SyntaxError rejection produces 0 reloads and the error goes straight to the error boundary, while `isChunkLoadError(err, "anything.js", /.*/)` returns true for the same error.

The consequence is an API-contract one rather than a runtime one: a consumer with non-default chunk naming passes a custom pattern here, sees a parameter that reads as supported, and gets no behavior change or warning.

It is benign at runtime for the reason already established on the Firefox-wording thread: webpack tags every `import()` chunk failure with `error.name = "ChunkLoadError"` (JsonpChunkLoadingRuntimeModule.js:186-191, including the "200 OK but the chunk never registered" case), so `lazyWithReload` never needs the filename heuristic at all.

Suggested fix: drop the parameter from the signature and from the call at :175, or keep it and state in the header comment (:166-170) that a custom pattern only applies to `initChunkErrorRecovery`.

smarcet (changes requested):
@santipalenque please review

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+  filename,
+  chunkFilenamePattern = DEFAULT_CHUNK_FILENAME_PATTERN
+) => {
+  const message = (error && error.message) || (typeof error === "string" ? error : "");
+  // error.name is set when available; otherwise fall back to the word
+  // "SyntaxError" in the flat message string (window.onerror without event.error).
+  const looksLikeSyntaxError =
+    (error && error.name === "SyntaxError") || /SyntaxError/i.test(message);
+  if (!looksLikeSyntaxError) return false;
+  return Boolean(filename) && chunkFilenamePattern.test(filename);
```
smarcet: @santipalenque The filename cross-check on this line does not discriminate in a production build, so any uncaught SyntaxError thrown from app code triggers a full page reload.

`DEFAULT_CHUNK_FILENAME_PATTERN` (line 34) is meant to prove a filename belongs to a lazy chunk, but summit-admin emits its *entry* bundles with the same naming: `output.filename: "[name]_[contenthash].js"` and `chunkFilename: "[name]_[contenthash].js"` (summit-admin `webpack.prod.js:11-12`). Every frame filename in the app therefore matches the pattern, and `initChunkErrorRecovery`'s `onError` (line 204) passes exactly that filename straight into `reloadOnChunkError`.

Concrete failure: summit-admin `src/pages/emails/edit-email-template-page.js:86` runs `JSON.parse(json_preview)` unguarded on admin-typed JSON, wired to `<button onClick={this.handlePopupClose}>` (`:199`) over an editable JSON editor (`:182-185`). React rethrows event-handler errors rather than routing them to a boundary (`react-dom/cjs/react-dom.development.js:322-329`, `rethrowCaughtError`), so a typo in the template JSON surfaces as an uncaught SyntaxError whose `filename` is the content-hashed bundle. I executed this against the module: 1 `window.location.reload()`, 1 `preventDefault()`. The admin loses the unsaved template and never sees an error — the page just reloads.

Suggested fix: delete the SyntaxError heuristic. Webpack's JSONP runtime already covers the stale-chunk-serving-HTML case it was written for — on `event.type === 'load'` (script 200 OK but the chunk never registered) it sets `errorType = 'missing'` and `error.name = 'ChunkLoadError'` (`webpack/lib/web/JsonpChunkLoadingRuntimeModule.js:186-191`), which `isWebpackChunkLoadError` already matches. If you keep the branch, it needs a discriminator for script *parse-time* failures — a filename shape the entry bundles also satisfy is not one.

This is the half of the earlier "filename-less SyntaxError branch" thread that did not land: requiring a filename fixed the `unhandledrejection` path (filename `undefined`), but not the global `error` path.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+
+export const chunkErrorSentryBeforeSend = (event) => {
+  const exceptionValues = event.exception?.values || [];
+  const matchesChunkError = exceptionValues.some((exceptionValue) =>
+    isChunkLoadError(
+      { name: exceptionValue.type, message: exceptionValue.value },
+      getExceptionFilename(exceptionValue)
+    )
+  );
+  return matchesChunkError && isChunkReloadPending() ? null : event;
```
smarcet: @santipalenque This filter is inert in the expected bootstrap order — the first chunk error, the one it exists to suppress, reaches Sentry anyway.

`@sentry/core` installs its global hooks by *property assignment*, not `addEventListener`: `GLOBAL_OBJ.onunhandledrejection = ...` (`build/esm/utils-hoist/instrument/globalUnhandledRejection.js:21-35`) and `GLOBAL_OBJ.onerror = ...` (`.../globalError.js:19-45`), with the source comment stating this is deliberate. So whoever registers first runs first. With the natural wiring — `Sentry.init()` at the top of bootstrap, `initChunkErrorRecovery()` after — Sentry captures and runs `beforeSend` *before* uicore's listener sets `reloadTriggeredThisPage`.

Measured both orders against this module with a stubbed Sentry hook:

```
Sentry.init() first             -> pending=false at capture -> KEPT (reload still happens)
initChunkErrorRecovery() first  -> pending=true  at capture -> DROPPED
```

So `isChunkReloadPending()` is false exactly when it matters, and the filter only ever acts on the second-and-later chunk errors in the same tick. The order dependency is not documented anywhere.

Suggested fix: document that `initChunkErrorRecovery()` must be called *before* `Sentry.init()`, and state it in the module header next to the `beforeSend` contract. One caveat that has to be verified before shipping it: in that order, a chunk error landing in the window between the two calls would fire `reportRecoveryFailed` while Sentry has no client yet, and that report is lost silently. Un-gating the filter from `pending` is not an alternative — the gate is what keeps unrecovered chunk errors visible, per the bare-`React.lazy` thread.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+export const reloadOnChunkError = (error, filename, chunkFilenamePattern) => {
+  if (!isChunkLoadError(error, filename, chunkFilenamePattern)) return false;
+
+  if (reloadTriggeredThisPage) return false;
+
+  const currentBuild = getBuildFingerprint();
+  const state = readReloadState();
+
+  if (state && state.build === currentBuild) {
+    reportRecoveryFailed(error, filename);
```
smarcet: @santipalenque This report has no once-per-build guard, so a post-reload page load files one Sentry error event per failing chunk, and repeats on every subsequent load while the build stays stale.

`reloadTriggeredThisPage` (line 126) is only set on the reload path at line 140, so it dedups the *pre*-reload tick and nothing else. On the load where the build is still stale — the scenario this branch exists for — every chunk error takes this branch independently. Executed against the module: three chunk errors on a post-reload module instance produced three `Sentry.captureMessage` calls.

This is the same shape as the concurrent-failure thread that was fixed earlier; that fix just stopped one tick short of the branch that actually reports.

Suggested fix: persist a `reported: true` flag alongside the build in the reload state written at line 136, and skip `reportRecoveryFailed` when it is already set.

src/utils/__tests__/lazy-with-reload.test.js
```diff
@@ -0,0 +1,482 @@
...
+        expect(window.location.reload).not.toHaveBeenCalled();
+        expect(mockCaptureMessage).not.toHaveBeenCalled();
+    });
+
+    // Reproduces the "two deploys in the same long-lived tab" scenario: a
+    // reload already happened once for build A, and now build B (a distinct
+    // set of statically-loaded <script> tags) has its own, unrelated chunk
+    // failure - it should get its own fresh reload attempt rather than
+    // being silently blocked by the earlier, unrelated recovery.
+    test("allows a fresh reload when a chunk error occurs on a different build than the one already reloaded for", () => {
```
smarcet: @santipalenque The storage-unavailable path requested on two separate threads has no test. Both threads were resolved; the code landed, the test did not.

The behaviour is correct today — I verified it with a `sessionStorage` stub whose `setItem` throws: `handled=false`, 0 reloads, 1 `reportRecoveryFailed`, and `isChunkReloadPending()` stays false. That last part is the fragile bit: it depends on `reloadTriggeredThisPage = true` sitting at `lazy-with-reload.js:140`, *after* the `writeReloadState` guard at `:136-139`. Any future reordering silently restores the "permanent Suspense fallback, no reload, chunk exceptions dropped" behaviour, and nothing in the suite would catch it.

Suggested fix: add one test in this describe block that stubs `window.sessionStorage` with a throwing `setItem` and asserts no `window.location.reload()` call, one `reportRecoveryFailed`, `isChunkReloadPending() === false`, and that `lazyWithReload`'s promise rejects rather than hanging.

src/utils/lazy-with-reload.js
```diff
@@ -0,0 +1,240 @@
...
+  window.location.reload();
+  return true;
+};
+
+// Logs (once) that a prior chunk-load-error reload actually took effect, so
+// it's easy to confirm in the console that this recovery path fired.
+const confirmReloadIfPending = () => {
+  try {
+    if (window.sessionStorage.getItem(PENDING_CONFIRMATION_KEY) === "1") {
+      window.sessionStorage.removeItem(PENDING_CONFIRMATION_KEY);
```
smarcet: @santipalenque `RELOAD_STATE_KEY` is never cleared — this line clears only `PENDING_CONFIRMATION_KEY` — so after a recovery that actually worked, a later unrelated chunk failure on the same build gets no reload and files a report claiming the reload did not help.

The key is written at line 83 and read at line 74; there is no `removeItem` for it anywhere in the module. Failure scenario: a transient CDN/network blip fails a chunk, the reload fixes it, and the same build comes back — so the stored fingerprint now equals the current one. A second blip later in that session hits the `state.build === currentBuild` branch at `:131`, skips the reload, and reports "Chunk load error persisted after auto-reload" for a recovery that succeeded.

Suggested fix needs care: clearing the key here would reopen the stale-`index.html` reload loop the guard exists to stop, since bootstrapping successfully does not prove the chunks are fine. Record recovery only once a lazy chunk actually loads after the reload, and treat a state marked recovered as reload-eligible again.

smarcet (changes requested):
@santipalenque LGTM

src/utils/lazy-with-reload.js
smarcet, line 47: @santipalenque This filename check doesn't discriminate in a production build: summit-admin's entry bundles use the same `[name]_[contenthash].js` naming as chunks (`webpack.prod.js:11-12`), so *any* uncaught SyntaxError from app code matches and triggers a reload — e.g. a bad `JSON.parse` in an `onClick` handler reloads the page and eats the user's unsaved input.

Simplest fix is to drop the SyntaxError heuristic entirely: webpack already sets `error.name = "ChunkLoadError"` for the stale-chunk-serving-HTML case (`JsonpChunkLoadingRuntimeModule.js:186-191`), which `isWebpackChunkLoadError` catches. That also removes the need for `chunkFilenamePattern` across the module.

smarcet (approved):
LGTM

PR #303, opened 2026-07-31
Fix: use different query for add-on ddl

santipalenque:
as part of https://app.clickup.com/t/9014802374/86bat8h0n

## Summary by CodeRabbit

- **New Features**
  - Added generalized add-on type selectors for standard and Formik-integrated forms.
  - Add-on options now display clear names rather than raw result data.
  - Updated data retrieval to use the add-on types endpoint.
  - Updated selector entry points and release version.

- **Breaking Changes**
  - Replaced the former summit-specific selector exports with generalized add-on type selector names.


src/utils/query-actions.js
```diff
@@ -635,7 +635,7 @@ export const querySummitAddons = async (
     callback
 ) => {
     const endpoint = URI(
-        `/api/v1/summits/${summitId}/add-ons/metadata`
+        `/api/v1/summits/all/add-on-types`
```
smarcet, line 637: @santipalenque This endpoint switch drops the alphabetical ordering the previous one guaranteed, and nothing replaces it.

`add-ons/metadata` sorted server-side: `DoctrineSummitSponsorshipAddOnRepository::getMetadata()` runs `SELECT t.name FROM SummitSponsorshipAddOnType t ORDER BY t.name ASC`, so the client never had to ask for it. `add-on-types` is a generic paginated list — `GetAll::getAll()` leaves `$order = null` unless an `order` query param arrives, and `DoctrineRepository::getAllByPage()` applies `ORDER BY` only when `$order` is non-null. The dropdown will render in insertion/DB order: the four system-defined types first, then custom ones by creation order.

This becomes user-visible in summit-admin. Its current local `SummitAddonSelect` still hits the old endpoint and therefore shows a sorted list, so swapping to this component on `fix/replace-mui-formik-for-uicore` is a straight regression for the admin.

21 of the query functions in this same file already set `order`; `querySponsorshipsBySummit` (line 666) does exactly this against a comparable summit-api list endpoint. The endpoint whitelists the field — `OAuth2SummitSponsorshipAddOnTypesApiController::getOrderRules()` returns `['id', 'name']`, and the repository maps `'name' => 'e.name'`.

Fix: add `endpoint.addQuery("order", "+name");` after the `per_page` line.

src/utils/query-actions.js
```diff
@@ -635,7 +635,7 @@ export const querySummitAddons = async (
...
     );
     endpoint.addQuery("page", 1);
     endpoint.addQuery("per_page", MAX_PER_PAGE);
```
smarcet, line 640: @santipalenque `page`/`per_page` were inert against the old endpoint but take effect here, so this line silently caps the option list at 100.

`getMetadata()` returned the complete list — `array_column($rows, 'name')`, no LIMIT — and ignored both params. `add-on-types` genuinely paginates, and `_fetchPublic` hands the callback only `json.data` (line 31), so `total` and `last_page` are discarded. Add-on types are a **global** catalog (`SummitSponsorshipAddOnType` has no summit association), so the table accumulates across every show; once it passes 100 rows, types vanish from the dropdown with no signal to the admin and no way to select them.

This is the exact failure mode ratified as unacceptable for picker option lists in the FN vault — `sds/dropbox-materializer-participating-types-allowlist.md`, "Pagination completeness": *"a truncated list would silently hide types from the multi-select and admins would have no way to know they're missing… the spec must not rely on that implicitly."*

If the catalog is expected to stay small, raise the page size — summit-api allows `per_page` up to 500 (`PaginationValidationRules::PerPageMax`). If it can grow past that, follow the vault pattern: read `last_page` from page 1 and fan out the remaining pages before invoking the callback.

src/components/mui/addon-type-select.js
```diff
@@ -64,11 +64,11 @@ const SummitAddonSelect = ({
   );
 };
 
-SummitAddonSelect.propTypes = {
+AddonTypeSelect.propTypes = {
   value: PropTypes.string,
   summitId: PropTypes.number.isRequired,
```
smarcet: @santipalenque `summitId` no longer does anything but is still a required prop, here and on the Formik wrapper.

The query it feeds now targets `/api/v1/summits/all/add-on-types`, which is summit-agnostic, so the value is accepted and dropped. Two costs: every consumer keeps threading a meaningless prop (`manage-tier-addons-popup.js:229,326` on summit-admin's `fix/replace-mui-formik-for-uicore` still passes it), and a *required* prop tells the next reader the list is summit-scoped when it is a global catalog — the component was renamed to `AddonTypeSelect` precisely to drop the summit framing, so the prop now contradicts the rename.

Copilot's thread on `query-actions.js:636` covers the query function's parameter; this is the component-side half of the same cleanup. Suggest removing `summitId` from `AddonTypeSelect.propTypes`, from `MuiFormikAddonTypeSelect.propTypes` and from the `querySummitAddons` signature in one pass, coordinated with the summit-admin call sites — otherwise the dead prop outlives the rename.

src/components/mui/__tests__/addon-type-select.test.js
```diff
@@ -33,12 +36,12 @@ describe("SummitAddonSelect", () => {
   beforeEach(() => jest.clearAllMocks());
 
   test("renders a select combobox", () => {
-    const { container } = render(<SummitAddonSelect {...defaultProps} />);
+    const { container } = render(<AddonTypeSelect {...defaultProps} />);
     expect(container.querySelector("[role='combobox']")).toBeInTheDocument();
   });
 
   test("calls querySummitAddons on mount with summitId", async () => {
```
smarcet: @santipalenque The endpoint swap — the highest-risk line in this PR — is covered by no test, and this assertion pins a contract the implementation no longer honours.

`query-actions.js` has no test file at all (`src/utils/__tests__/` holds only `actions`, `external-store`, `money`, `questions_set`), and all 8 component tests that touch the module call `jest.mock` on it. Nothing exercises the URL, the query params, or the `json.data` unwrapping. That gap is why the original defect reached `main`: `getMetadata` returns a bare array (`ok()` → `response2XX()` → `Response::json($data)`, no envelope), so `_fetchPublic`'s `callback(json.data)` passed `undefined` into `results.map()` and the component threw. A `query-actions` test asserting the built request and the callback payload would have caught it — and would guard the new paginated shape, where the envelope does carry `data` (`PagingResponse::toArray()`).

Separately, line 48 asserts `toHaveBeenCalledWith(5, expect.any(Function))`, locking in a `summitId` argument the query now ignores — an implementation-detail assertion that outlived its behaviour.

Suggest adding `src/utils/__tests__/query-actions.test.js` that mocks `fetch` + `getAccessToken` and asserts the URL and params `querySummitAddons` builds plus the array the callback receives from a paginated envelope, and dropping the `summitId` argument from this assertion. (CodeRabbit's open thread already asks for the `r.name` option-mapping assertion — not duplicating it here.)

src/utils/query-actions.js
```diff
@@ -635,7 +635,7 @@ export const querySummitAddons = async (
     callback
 ) => {
```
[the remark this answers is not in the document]
smarcet: The dead parameter is real, but two things here don't hold up. There's no `no-unused-vars` risk — this repo has no ESLint config at all (no `.eslintrc*`, no `eslint` key in `package.json`). And a dual call style would be adding runtime type detection to a public API for zero legacy callers: `querySummitAddons` has exactly one caller in the ecosystem (`addon-type-select.js:29`), since summit-admin uses its own local copy in `sponsor-actions.js`. The only `typeof` guards in this file (lines 30, 71) check callback callability, not argument position — no precedent for an overloaded signature.

Suggest just deleting the parameter, together with the `summitId` propTypes on both components and the summit-admin call sites. Details in the thread on `addon-type-select.js:69`.

src/components/index.js
```diff
@@ -68,7 +68,7 @@ export {default as MuiMenuButton} from './mui/menu-button'
 export {default as MuiSearchInput} from './mui/search-input'
 export {default as MuiShowConfirmDialog} from './mui/showConfirmDialog'
 export {default as MuiSponsorAddonSelect} from './mui/sponsor-addon-select'
-export {default as MuiSummitAddonSelect} from './mui/summit-addon-select'
+export {default as AddonTypeSelect} from './mui/addon-type-select'
```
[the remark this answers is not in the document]
smarcet: Agreed on the inconsistency — `AddonTypeSelect` is the only export in the MUI block without the prefix, and this same PR adds `MuiFormikAddonTypeSelect` with it, so the two halves disagree with each other.

An alias would be worse than the problem though: two public names for one component, both of which then have to be maintained and deprecated. A straight rename to `MuiAddonTypeSelect` is free right now — nothing imports it from the barrel yet, since the consumer on summit-admin's `fix/replace-mui-formik-for-uicore` uses the direct path `lib/components/mui/formik-inputs/addon-type-select`. Leaving it costs a breaking rename later.

src/utils/query-actions.js
santipalenque: I think 100 add-on types is more than enough @smarcet . If it ever grows pass that we would need to refactor this into a async select instead.,

src/components/mui/__tests__/addon-type-select.test.js
santipalenque: dropping the summitId in tests, but don't think we need a test for pagination payload/reponse, this is the standard in our API.

smarcet (approved):
LGTM

